PR #1, opened 2015-11-05
Feature/svm grouping

mktal:
Add grouping support for SVM regression

Issue: https://issues.apache.org/jira/browse/MADLIB-913

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,8 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     if(APPLE)
         set(CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem ")
     endif(APPLE)
+elseif(CMAKE_C_COMPILER_ID MATCHES "Clang")
+    set(CMAKE_CXX_FLAGS "-stdlib=libstdc++")
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
 # force a `m4_' prefix to all builtins
--- a/src/modules/convex/linear_svm_igd.cpp
+++ b/src/modules/convex/linear_svm_igd.cpp
@@ -53,12 +53,9 @@ linear_svm_igd_transition::run(AnyType &args) {
     // For other tuples: args[0] holds the computation state until last tuple
     GLMIGDState<MutableArrayHandle<double> > state = args[0];
 
-    const double lambda = args[6].getAs<double>();
-    const bool isL2 = args[7].getAs<bool>();
-    const int nTuples = args[8].getAs<int>();
-
-    // initilize the state if first tuple
+    // initialize the state if first tuple
     if (state.algo.numRows == 0) {
+        LinearSVM<GLMModel, GLMTuple >::epsilon = args[9].getAs<double>();;
         if (!args[3].isNull()) {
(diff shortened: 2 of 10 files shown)


src/modules/convex/linear_svm_igd.cpp
```diff
@@ -95,19 +90,16 @@ linear_svm_igd_transition::run(AnyType &args) {
...
     // apply IGD with regularization
-    if (isL2) {
-        L2<GLMModel>::scaling(state.algo.incrModel, lambda, nTuples, state.task.stepsize);
-        LinearSVMIGDAlgorithm::transition(state, tuple);
-    } else {
-        LinearSVMIGDAlgorithm::transition(state, tuple);
-        L1<GLMModel>::clipping(state.algo.incrModel, lambda, nTuples, state.task.stepsize);
-    }
-    // objective function and its gradient
+    L2<GLMModel>::scaling(state.task.model, state.algo.incrModel, state.task.stepsize);
```
iyerr3: Why not check the isL2 variable and only perform the necessary computation?

src/modules/convex/linear_svm_igd.cpp
```diff
@@ -152,9 +144,18 @@ linear_svm_igd_final::run(AnyType &args) {
 
     // Aggregates that haven't seen any data just return Null.
     if (state.algo.numRows == 0) { return Null(); }
+
```
iyerr3: Same as above - I would prefer to have one `if (isL2)` that decides between the L2 and L1 functions.

src/modules/convex/linear_svm_igd.cpp
mktal: The L2 and L1 regularization are not exclusive. We can perform both of them by setting the corresponding regularization parameters to non-zero.

src/ports/postgres/modules/utilities/in_mem_group_control.py_in
```diff
@@ -7,6 +7,141 @@
...
+            return failed_grp_keys
+        # initialize state to None
+        if col_grp_state == '':
+            self._is_none = True
+            for s in ret_states:
+                self._state[s[col_grp_key]] = None
+            return failed_grp_keys
+        for t in ret_states:
+            _grp_key, _grp_state = t[col_grp_key], t[col_grp_state]
+            if _grp_state is None: failed_grp_keys.append(_grp_key)
```
iyerr3: Let's put the statement in a separate line per [Pep 8](https://www.python.org/dev/peps/pep-0008/#other-recommendations)

src/modules/convex/linear_svm_igd.cpp
mktal: For now the users can only choose either l1 or l2. We can support both in the future if necessary

CMakeLists.txt
```diff
@@ -103,6 +103,8 @@ if(CMAKE_COMPILER_IS_GNUCXX)
     if(APPLE)
         set(CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem ")
     endif(APPLE)
+elseif(CMAKE_C_COMPILER_ID MATCHES "Clang")
```
iyerr3: @mktal let's move this out as a separate JIRA+PR since it's not related to SVM.

PR #4, opened 2015-11-21
SVM: Add cross validation support and generic CrossValidator class

mktal:
JIRA: MADLIB-915

Authors:
Xiaocheng Tang <xiaocheng.t@gmail.com>
Rahul Iyer <rahiyer@gmail.com>

Changes:
- Add cross validation support on lambda, epsilon, init_stepsize, max_iter, and decay_factor
- Add support for optionally writing validation results to a sql table
- Add support for lazy-generation of cv datasets
- Add internal generic CrossValidator class which is used for implementing this issue
- Refactoring SVM for better modularity
- Ignore cv on epsilon when it is classification
- Cross validation now works when independent variables are queries
- Fixed "zero length field name in format" error in python < 2.7
- Fix error messages so that it pass the svm input tests

### DIFF
--- a/src/ports/postgres/modules/svm/svm.py_in
+++ b/src/ports/postgres/modules/svm/svm.py_in
@@ -326,15 +326,12 @@ def _svm_parsed_params(schema_madlib, source_table, model_table,
     """
     Executes the linear support vector classification algorithm.
     """
-    grouping_str = _verify_grouping(schema_madlib,
-                                    source_table,
-                                    grouping_col)
+    grouping_str = _verify_grouping(schema_madlib, source_table, grouping_col)
 
     kernel_func = _verify_kernel(kernel_func)
 
     # arguments for iterating
-    n_features = num_features(source_table,
-                              independent_varname)
+    n_features = num_features(source_table, independent_varname)
 
     args = {
             'rel_args': unique_string(desp='rel_args'),
--- a/src/ports/postgres/modules/svm/svm.py_in
+++ b/src/ports/postgres/modules/svm/svm.py_in
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import plpy
 
 from utilities.control import MinWarning
@@ -9,12 +11,21 @@ from utilities.utilities import add_postfix
 from utilities.utilities import _string_to_array_with_quotes
 from utilities.utilities import _string_to_array
 from utilities.utilities import _assert
+from utilities.utilities import num_features
 
 from utilities.validate_args import cols_in_tbl_valid
 from utilities.validate_args import input_tbl_valid
 from utilities.validate_args import output_tbl_valid
 from utilities.validate_args import is_var_valid
 from utilities.validate_args import get_expr_type
+
+from validation.internal.cross_validation import CrossValidator
+
+from collections import namedtuple
+from functools import partial
+from operator import itemgetter
+from operator import attrgetter
+from itertools import product, repeat, imap
 
 
 def _compute_svm(args):
@@ -23,6 +34,7 @@ def _compute_svm(args):
 
     @return Number of iterations that has been run
     """
+    args['stepsize'] = args['init_stepsize']
     iterationCtrl = GroupIterationController(args)
     with iterationCtrl as it:
         it.iteration = 0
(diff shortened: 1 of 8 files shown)


src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -66,211 +253,118 @@ def svm(schema_madlib, source_table, model_table,
...
+    args.update(_verify_params_dict(params_dict))
+    args.update(_process_epsilon(is_svc, args))
+    args.update(_svc_or_svr(is_svc, source_table, dependent_varname))
+
+    # place holder for compatibility
+    plpy.execute("CREATE TABLE pg_temp.{0} AS SELECT 1".format(args['rel_args']))
+    # actual iterative algorithm computation
+    n_iters_run = _compute_svm(args)
+    _summary(n_iters_run, model_table, args)
+>>>>>>> b105d1c... SVM: Add cross validation support and generic CrossValidator class
```
iyerr3: @mktal Invalid line. Also run the py files through a linter. There are a few places where we're not following the pep8 guidelines.

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -389,11 +521,20 @@ def _process_epsilon(is_svc, args):
...
+        epsilon = 0.0
+        select_epsilon = '{epsilon}'.format(epsilon=epsilon)
+    elif not grouping_col or not eps_table:
         if eps_table:
-            plpy.warning('SVM: ignoring the input epsilon table!')
-        select_epsilon = str(epsilon)
+            plpy.warning('SVM warning: no grouping and '
+                         ' ignore the input epsilon table!')
+        # c code does SVC if epsilon is zero
+        if epsilon == 0: epsilon = 0.00001
```
iyerr3: I thought we don't need this after adding the is_svc boolean to the c code

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -440,55 +595,68 @@ def _process_epsilon(is_svc, args):
...
-    params_types = {'init_stepsize': float,
-                    'decay_factor': float,
-                    'max_iter': int,
-                    'tolerance': float,
-                    'lambda': list,
-                    'norm': str,
-                    'n_folds': int,
-                    'epsilon': float,
-                    'eps_table': str}
+    params_default = {
```
iyerr3: Not clear to me why 'init_stepsize', 'max_iter', and 'decay_factor' are lists? Do we expect to CV over these parameters?

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -440,55 +595,68 @@ def _process_epsilon(is_svc, args):
...
     # validate lambda
-    if hasattr(params_vals['lambda'], '__len__'):
-        # this check should be removed after cross validation is added
-        if len(params_vals['lambda']) != 1:
-            plpy.error("{0} error: lambda must be a scalar or of length 1 when n_folds is 0 or 1".format(module))
-        params_vals['lambda'] = params_vals['lambda'][0]
-    _assert(params_vals['lambda'] >= 0,
-            "SVM error: lambda must be non-negative!")
+    params_vals['lambda'] = map(float, params_vals['lambda'])
+    if [lmd for lmd in params_vals['lambda'] if lmd < 0]:
```
iyerr3: I would prefer to read this as `any(l < 0 for l in params_val['lambda']`

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -440,55 +595,68 @@ def _process_epsilon(is_svc, args):
...
+                        'n_folds': int,
+                        'validation_result': str,
+                        'epsilon': list,
+                        'eps_table': str}
 
     params_vals = extract_keyvalue_params(params,
                                           params_types,
                                           params_default)
-
     if params_vals['n_folds'] < 0:
```
iyerr3: let's use the `_assert` format to maintain consistency. I would like to add another PR to change all `_assert` to `assert_`

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -440,55 +595,68 @@ def _process_epsilon(is_svc, args):
...
-    _assert(params_vals['max_iter'] > 0,
-            "SVM error: max_iter must be positive")
-    _assert(params_vals['tolerance'] >= 0,
-            "SVM error: tolerance must be non-negative")
-    _assert(params_vals['epsilon'] >= 0,
-            "SVM error: epsilon cannot be less than 0")
+    if params_vals['norm'] != 'l1' and params_vals['norm'] != 'l2':
+        plpy.error("{0} error: norm must be either L1 or L2!".format(module))
+    if params_vals['tolerance'] < 0:
+        plpy.error("{0} error: tolerance must be non-negative!".format(module))
```
iyerr3: let's use the `assert` formulation for all the above error messages.

src/ports/postgres/modules/svm/svm.py_in
mktal: Early stopping in optimization serves the similar purposes as regularizations do. Specifying those optimization parameters as lists allows users to CV over them so that they are chosen to provide the `best` generalization ability while avoiding over- or under-fitting. Another practical concern is to make it easier for users to choose a proper init_stepsize which varies from case to case depending on both the model and the data.

src/ports/postgres/modules/svm/svm.py_in
mktal: This is so embarrassing....I should have at least run install-check after cherry-picking!

iyerr3: No problem - this has happened to me before :)

src/ports/postgres/modules/svm/svm.py_in
iyerr3: Also please look at the setting epsilon to low value part. I had added the `is_svc` boolean during the regression commit to avoid this.

PR #10, opened 2015-12-15
SVM: Add Gaussian kernel feature mapping

mktal:


methods/array_ops/src/pg_gp/array_ops.sql_in
```diff
@@ -481,6 +481,18 @@ m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `NO SQL', `');
  * @brief This function takes an array as the input and executes element-wise multiplication by the scalar provided as the second argument, returning the resulting array. It requires that all the values are NON-NULL. Return type is the same as the input type.
  * @param x Array x
  * @param k Scalar
+ * @returns Array with each element of cos(x) multiplied by scalar.
```
haying: The function signature below does not seem to allow scalar k as the comment says :-)

mktal: Thanks for pointing it out! This was however a 'draft' pull request.;-) And we already fix that in the latest version which have not been pushed yet. We also made quite a lot other changes as well.

methods/array_ops/src/pg_gp/array_ops.c
```diff
@@ -824,6 +836,25 @@ array_fill(PG_FUNCTION_ARGS){
...
+array_cos(PG_FUNCTION_ARGS){
+    if (PG_ARGISNULL(0)) { PG_RETURN_NULL(); }
+
+    ArrayType *v1 = PG_GETARG_ARRAYTYPE_P(0);
+    Oid element_type = ARR_ELEMTYPE(v1);
+    Datum v2 = float8_datum_cast(0, element_type);
+
+    ArrayType *res = General_Array_to_Array(v1, v2, element_cos);
+
+    PG_FREE_IF_COPY(v1, 0);
```
cwelton: This is a weird use of PG_FREE_IF_COPY that simply looks wrong to me.  This call is equivalent to pfree(v1), and since that is a passed in argument you are freeing something that does not belong to this function.

A correct call would be PG_FREE_IF_COPY(v1, PG_GETARG_ARRAYTYPE_P(0)), except that will never free anything so would be a no-op.  Ultimately this line of code should simply be removed.

cwelton:
-1 from me as the code currently stands. 

Freeing memory passed to a function can lead to instability of the database system and is a complete blocker for merge.

methods/array_ops/src/pg_gp/array_ops.c
iyerr3: So during the detoast, PG_GETARG_ARRAYTYPE_P does not create a copy? I was under the impression that we have to free that pointer since a copy is always created. All array ops functions in MADlib perform that free, based on similar functions in pg source code. 
If that's wrong then we'll have to make a pretty big change in our array_ops. 

Snippet from /src/backend/utils/adt/arrayfuncs.c: 
```
Datum
array_eq(PG_FUNCTION_ARGS)
{
	ArrayType  *array1 = PG_GETARG_ARRAYTYPE_P(0);
	ArrayType  *array2 = PG_GETARG_ARRAYTYPE_P(1);
	Oid			collation = PG_GET_COLLATION();
	int			ndims1 = ARR_NDIM(array1);
	int			ndims2 = ARR_NDIM(array2);
	int		   *dims1 = ARR_DIMS(array1);
	int		   *dims2 = ARR_DIMS(array2);
...
...
...
	/* Avoid leaking memory when handed toasted input. */
	PG_FREE_IF_COPY(array1, 0);
	PG_FREE_IF_COPY(array2, 1);

	PG_RETURN_BOOL(result);
}
```

cwelton: Hmm, taking a closer look at the PG_FREE_IF_COPY macro I see I misread how it examines the second parameter.

Ignore my previous comment, this looks safe.

cwelton: In answer to your question, you can think of the parameter that is being received as a union between (toastid, text_pointer).  If you receive a pointer then you do not get a copy, if you get a toastid then the GET_ARRAY_TYPE function will detoast it and return a pointer to you.  In neither case do you "get a copy" of a pointer that was passed as input, which is what led to the initial confusion here - "free_if_copy" is a misleading name for the macro.

methods/array_ops/src/pg_gp/array_ops.c
```diff
@@ -1929,3 +1971,126 @@ General_Array_to_Array(
     return pgarray;
 }
 
+/******* internal functions ********/
```
iyerr3: This function would be much easier to code and maintain if we used the abstraction layer + eigen to compute the matrix dot product. See closestColumnsAndDistances in metric.cpp for an example of how a matrix and a vector are used in an operation.

iyerr3: Turns out we have a similar function in svd.cpp (svd_vec_mult_matrix). we can build something similar - it should end up in matrix_ops.cpp.

src/ports/postgres/modules/linalg/matrix_ops.py_in
```diff
@@ -3093,3 +3094,75 @@ def matrix_ones(schema_madlib, row_dim, col_dim, matrix_out, out_args):
...
+    _validate_output_table(matrix_out)
+    _assert(row_dim > 0 and col_dim > 0,
+            "Matrix error: Invalid dimensions for input matrices: {0} x {1}".
+            format(row_dim, col_dim))
+
+    out_args = parse_matrix_args(
+        out_args, in_default_args={'row': 'row', 'col': 'col', 'val': 'val'})
+
+    distribution = 'uniform' if not distribution else distribution.lower()
+    supported_dist = ['uniform', 'normal']
```
iyerr3: We have the same pattern in multiple places (here, kernel selection in svm, norm selection in matrix_ops) . If you're up for it, please move this to a function and update all locations. Else please create a JIRA to make this update.

src/ports/postgres/modules/svm/kernel_approximation.py_in
```diff
@@ -0,0 +1,481 @@
...
+                    select id as {rd_id}, val as {rd_val} from {data}
+                    where desp = 'weights';
+
+                drop view if exists {rd_offset};
+                create temp view {rd_offset} as
+                    select id as {rd_id}, val as {rd_val} from {data}
+                    where desp = 'offsets';
+                     """.format(**locals()))
+        params = cls.parse_params(kernel_params)
+        in_memory = params.pop('in_memory', True)
```
iyerr3: why pop instead of get?

mktal: params is later used to construct the kernel object which does not take the argument in_memory

methods/array_ops/src/pg_gp/array_ops.c
mktal: Thanks for the pointer. I agree that the current implementation is a bit too complicated since it relies on the low-level c layer.

src/ports/postgres/modules/svm/kernel_approximation.py_in
iyerr3: I would still recommend get() and add **kwargs in the derived class __init__(). Any time we use a **params in the caller, it's best to add a **kwargs in the definition. That way it won't fail if someone later adds another parameter in parse_params() and does not update the derived classes.

src/ports/postgres/modules/svm/kernel_approximation.py_in
```diff
@@ -0,0 +1,481 @@
...
+        self.original_table = dict(source_table=source_table,
+                                   independent_varname=independent_varname,
+                                   dependent_varname=dependent_varname)
+
+        schema_madlib = self.schema_madlib
+
+        grouping_col = ("NULL::integer as {0}"
+                        .format(unique_string(desp='grouping_col'))
+                        if not grouping_col
+                        else "{0}".format(grouping_col))
```
iyerr3: `str(grouping_col)` would be simpler than `"{0}".format(grouping_col)`

iyerr3: IMHO writing it as below is cleaner
```
if not grouping_col: 
    grouping_col = "NULL::integer as " + unique_string(desp='grouping_col')
```

src/ports/postgres/modules/svm/kernel_approximation.py_in
mktal: I understand. I used **kwargs in the base class but in the derived class I am trying to make the interface more restrictive such that no unexpected variables got passed to base class through, for example, super().__init__(**locals()). Plus, it will be more clear to the caller this way what are expected to create the object

src/ports/postgres/modules/svm/kernel_approximation.py_in
```diff
@@ -0,0 +1,481 @@
...
+                from (
+                    select
+                        {dependent_varname} as {target_col},
+                        {independent_varname} as {features_col},
+                        {id_col},
+                        {grouping_col}
+                    from {source_table}
+                ) q cross join (select {val} from {rd_weights}) as rw
+                    cross join (select {val} from {rd_offset}) as ro
+        """.format(id_col=id_col,
```
iyerr3: To avoid the long format list (in multiple places) I would use the actual variable names in the format specifier and just use `.format(**locals())`

methods/array_ops/src/pg_gp/array_ops.c
mktal: Just pushed the changes. Please take a look and comment.

PR #19, opened 2016-02-20
Doc updates for SVM

fmcquillan99:


fmcquillan99:
@mktal please review these doc updates and merge

PR #21, opened 2016-02-22
Path: Allow multiple matches in single partition

iyerr3:
JIRA: MADLIB-917

Path query was rewritten to use array functions. This was necessary
since multiple matches produced a set of matches for each group. This
set value has to be unnest - this was not possible in GPDB since the
GROUP BY placed the executor in a context that does not allow SRF.
Solution was to avoid the unnest by using array functions for all
computations and finally unnesting without the GROUP BY.


iyerr3:
@mktal Please review and merge.

PR #22, opened 2016-02-23
SVM: Add Polynomial Kernel

mktal:
JIRA: MADLIB-938

The features are transformed using Kar and Karnick's appoach, which applies to any positive definite
kernel where the kernel function admits a Maclaurin expansion. This commit implements the special case
when the kernel function is polynomial.

@orhankislal Please review and comment


mktal:
This should fix it.

mktal:
@iyerr3 Please review and merge

methods/array_ops/src/pg_gp/array_ops.c
```diff
@@ -702,8 +702,14 @@ array_sub(PG_FUNCTION_ARGS){
 PG_FUNCTION_INFO_V1(array_mult);
 Datum
 array_mult(PG_FUNCTION_ARGS){
-    if (PG_ARGISNULL(0)) { PG_RETURN_NULL(); }
-    if (PG_ARGISNULL(1)) { PG_RETURN_NULL(); }
+    if (PG_ARGISNULL(0) && PG_ARGISNULL(1)) { PG_RETURN_NULL(); }
```
iyerr3: Not clear why this change is needed - this is saying that NULL * X = X, which indicates that NULL is treated as identity. The database does not make that assumption.

mktal: This change is made so that array_mult can be used as SFUNC in UDA, e.g., madlib.prod() to make sure that the null state is initialized with the first tuple in the table. This is consistent with array_add that NULL + X = X

mktal:
Well, I agree that NULL being treated as identity is a little weird. In that sense it is not consistent with array_add

methods/array_ops/src/pg_gp/array_ops.c
mktal: Well, I agree that NULL being treated as identity is a little weird. In that sense it is not consistent with array_add

mktal: Generally I try to avoid treating NULL as `zero` in multiplication because it produces zero even though only one entry of your data is NULL. Having said that, in numpy `np.nan*2 = np.nan`. But  `np.nan + 2` also results in np.nan.

decibel: This is very much against the database view of how NULLs work. Almost anything you do with NULL should result in NULL.

mktal:
@iyerr3 Please review and merge

methods/array_ops/src/pg_gp/array_ops.c
iyerr3: Agree with @decibel . Since array_mult/prod is not being used in SVM, we can remove the change from here and have a separate discussion for it.

iyerr3:
I merged this (commit: 3e576c3) but forgot to close the PR in the commit message. 

@mktal Please close this.

PR #28, opened 2016-03-01
SVM: Using temp table when generating random matrices

mktal:


iyerr3:
Is the random matrix table not needed for prediction?

mktal:
After training those random matrices are gathered into a single table which is used in prediction. The random matrices then need to be garbage-collected.

iyerr3:
Judging by how 'table_type' is used, I would expect 'TEMP' is the only useful/valid value. We should change that parameter to a boolean and add the 'TEMP' text in the called function.

PR #29, opened 2016-03-11
Path: Return results for each match

iyerr3:
JIRA: MADLIB-973

- Previously path returned the result per partition. This work changes
  that to result per match within each partition.
- 'agg_func' converted to a default argument.
- Match ID and symbol string added to the output tuples table.

### DIFF
--- a/src/ports/postgres/modules/utilities/path.py_in
+++ b/src/ports/postgres/modules/utilities/path.py_in
@@ -28,22 +28,19 @@ import string
 import re
 
 from control import MinWarning
-from utilities import unique_string
-from utilities import _assert
-from utilities import add_postfix
+from utilities import unique_string, _assert, add_postfix
+from utilities import py_list_to_sql_string
 from validate_args import get_cols
-from validate_args import input_tbl_valid
-from validate_args import output_tbl_valid
+from validate_args import input_tbl_valid, output_tbl_valid, is_var_valid
 from validate_args import quote_ident
 # ------------------------------------------------------------------------
-
 
 m4_changequote(`<!', `!>')
 
 
 def path(schema_madlib, source_table, output_table, partition_expr,
          order_expr, pattern_expr, symbol_expr, agg_func,
-         persist_rows=False, **kwargs):
+         persist_rows=None, **kwargs):
     """
         Perform regular pattern matching over a sequence of rows.
 
@@ -64,18 +61,15 @@ def path(schema_madlib, source_table, output_table, partition_expr,
             as_partition_expr = "1 = 1 as all"
         else:
             as_partition_expr = partition_expr
+        if persist_rows is None:
+            # persist_rows = None implies no preference
+            # persist_rows = False implies do not store the matched rows
+            persist_rows = not bool(agg_func)
+
         _validate(source_table, output_table, partition_expr, order_expr,
                   pattern_expr, symbol_expr, agg_func, persist_rows)
 
-        # replace each occurence of the original symbol with the new
-        # perform this operation in descending order of length to avoid substituting
-        # subset of any symbol
-        sym_mapping, sym_str = _parse_symbol_str(symbol_expr)
-        old_sym_desc = list(reversed(sorted(sym_mapping.keys(), key=len)))
-        replace_pattern = re.compile('|'.join(old_sym_desc), re.IGNORECASE)
-        pattern_expr = replace_pattern.sub(
-            lambda m: sym_mapping[re.escape(string.lower(m.group(0)))],
-            pattern_expr)
+        new_pattern_expr, long_sym_str, short_sym_str = _parse_symbol_str(symbol_expr, pattern_expr)
 
         # build variables for intermediate objects
         input_with_id = unique_string('input_with_id')
@@ -83,8 +77,13 @@ def path(schema_madlib, source_table, output_table, partition_expr,
         id_col_name = unique_string('id_col')
         matched_partitions = unique_string('matched_partitions')
         seq_gen = unique_string('seq_gen')
-        symbol_name_str = unique_string('symbol')
-        all_input_cols_str = ', '.join(get_cols(source_table, schema_madlib))
+        short_sym_name_str = unique_string('short_sym')
+        long_sym_name_str = unique_string('long_sym')
+        match_to_row_id = unique_string('match_to_row_id')
+        match_id = unique_string('match_id')
+
+        all_input_cols = [i.strip() for i in get_cols(source_table, schema_madlib)]
+        all_input_cols_str = ', '.join(all_input_cols)
         if persist_rows:
             matched_rows = add_postfix(output_table, "_tuples")
             table_or_view = 'TABLE'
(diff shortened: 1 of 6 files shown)


iyerr3:
@fmcquillan99 Please pull this branch to review/update documentation. 

@mktal Please give this a review.

iyerr3:
@fmcquillan99 New commit should return the original symbols

iyerr3:
@fmcquillan99 - Added a new commit to ensure that the symbol order is maintained. The expected behavior is "if a row matches with two symbols, the symbol that comes **first** in the symbol definition list will be given higher priority". 

@mktal I have simplified the code and added more comments. Please give this a read and let me know if it's easier on the eye.

fmcquillan99:
@iyerr3 I tested the symbol ordering and it seems to work fine.  My other basic tests pass as well, so this feature seems to be fine.  I have completed the docs & examples & will send early next week.

I also asked a Data Scientist pal of mine to test it out to get another set of eyes on it.  Will attach his findings when he reports back

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -387,11 +395,16 @@ def _parse_symbol_str(symbol_expr, pattern_expr):
...
+    # build a case statement to search a tuple for each definition and pick the
+    # appropriate symbol.
+    orig_sym_case_stmt = []
+    new_sym_case_stmt = []
+    general_case_stmt = "WHEN {0} THEN '{1}'::text"
+    for k in orig_symbols_ordered:
+        orig_sym_case_stmt.append(general_case_stmt.format(orig_sym_definitions[k], k))
+        new_sym_case_stmt.append(general_case_stmt.format(orig_sym_definitions[k],
+                                                          old_to_new[k.lower()]))
+    return (new_pattern_expr, '\n'.join(orig_sym_case_stmt), '\n'.join(new_sym_case_stmt))
```
decibel: I believe the example and returns sections of the function header need to be updated.

iyerr3: Good catch. Updated.

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -118,140 +120,175 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
+            CREATE VIEW {length_of_matches} AS
+            SELECT
+                {p_col_name_str},
+                {match_to_row_id},
+                array_agg(length(matches) ORDER BY match_index) AS matches,
+                array_agg(match_index ORDER BY match_index) AS match_indices
+            FROM (
+                SELECT
+                    {p_col_name_str},
+                    {match_to_row_id},
```
mktal: Is it necessary to have {match_to_row_id} in both {length_of_matches} and {length_of_between_matches}?

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -118,140 +120,175 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
-                                                ) t1
-                                            GROUP BY {p_col_name_str}, matched_ids, sym_str
-                                            ) t2
-                                        ) subq2
-                                    GROUP BY {p_col_name_str}, matched_ids
-                                ) q2
+                                {length_of_matches} as q2
                                 USING ({p_col_name_str})
-                            GROUP BY {p_col_name_str}, q1.matched_ids, match_splits, matches
+                            GROUP BY {p_col_name_str}, q1.{match_to_row_id}, match_splits, matches, match_indices
```
mktal: why not just `group by {p_col_name_str}`?

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -118,140 +120,175 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
                      FROM (
+                        -- use cumulative sum to get the start and end of each match
+                        -- (counting from start of symbol string)
                         SELECT
                             {p_col_name_str},
-                            matched_ids,
-                            {m}.array_cum_sum({m}.array_add(match_splits, prev_matches)) AS left_range,
-                            {m}.array_cum_sum({m}.array_add(match_splits, matches)) AS right_range
+                            {match_to_row_id},
+                            {m}.array_cum_sum({m}.array_add(match_splits, prev_matches)) AS match_start,
```
mktal: +1

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -118,140 +120,175 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
+                            {m}.array_cum_sum({m}.array_add(match_splits, matches)) AS match_end,
```
mktal: +1

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -108,8 +107,11 @@ def path(schema_madlib, source_table, output_table, partition_expr,
                             *,
                             nextval('{seq_gen}') AS {id_col_name},
```
mktal: why not simply `row_number() over ()`? (no need to create and drop the sequence)

src/ports/postgres/modules/utilities/path.py_in
iyerr3: yep - don't need it both views, since only 1 is being used downstream.

src/ports/postgres/modules/utilities/path.py_in
iyerr3: actually the `group by` is an artifact of the old query. After the refactor, we don't need grouping here. good catch.

src/ports/postgres/modules/utilities/path.py_in
iyerr3: the sequence should be more efficient in this case. We only need a unique id for each row, it doesn't have to be gapless - so the sequence would be better.

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -118,140 +120,172 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
+                        SELECT
+                            {p_col_name_str},
+                            {match_to_row_id},
+                            sym_str,
+                            (regexp_matches(
+                                   sym_str, '(?i)({new_pattern_expr})', 'g'))[1] AS matches
+                        FROM
+                           {matched_partitions}
+                        ) t1
+                    GROUP BY {p_col_name_str}, {match_to_row_id}, sym_str
```
mktal: I believe this can also be simplified to `group by {p_col_name_str}`

iyerr3: no it can't. the count(...) is the only aggregate there and we need the other columns in subsequent outer query.

src/ports/postgres/modules/utilities/path.py_in
decibel: What are you basing that on? I don't believe there's any significant performance benefit to sequences over row_number() (actually, a quick glance at the code leads me to think sequences will be slower). On top of that, creating a sequence adds to catalog bloat (not to mention the cost of creating and then dropping the sequence).

iyerr3: you're right. doing a quick test: 
```
select count(*) from (select nextval('seq_test') as id from generate_series(1, 10000000)) q;
 Time: 11975.435 ms

select count(*) from (select row_number() over() as id from generate_series(1, 10000000)) q;
Time: 5716.938 ms
```

PR #31, opened 2016-03-11
Elastic Net: Skip arrays with NULL values in predict

njayaram2:
Jira: MADLIB-919

Having NULL values in the input array led to an error while converting the array to a MappedColumnVector. This fix now skips prediction for arrays with NULL values. This also includes new test cases to the install check of elastic net.

@iyerr3


iyerr3:
General comment: Try to keep the commit title under 50 chars and definitely keep the body text under 72 chars. See Tim Pope's blog for [good commit messaging habits.](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)

PR #32, opened 2016-03-15
Recursive_partitioning: Fix random forrest error message

orhankislal:
JIRA: MADLIB-975
Random forest train gave a non-descript error if every data point had a NULL feature. Added an if check and and assert to give a proper description of the situation.


src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
```diff
@@ -669,6 +669,8 @@ def _get_n_and_deplist(training_table_name, dependent_variable, filter_null):
                    source=training_table_name,
                    filter_null=filter_null)
     r = plpy.execute(sql)[0]
+    if r['n_rows'] is None:
+        return None, None
```
iyerr3: It's better to return `0` instead of the `None` for `n_rows`. The _assert in the calling function could be changed to `n_rows > 0`

PR #35, opened 2016-04-06
        Elastic Net: Skip arrays with NULL values in train

njayaram2:
Jira: MADLIB-978

        Having NULL values in the input array of the training
        data was leading to an unhandled exception. This fix
        now catches the exception and skips such input arrays.
        The fix also modifies the code which was used to normalize
        the input data (independent variable), which now ignores
        such arrays with NULLs while normalizing.
        The number of rows in the input table was used while
        normalizing the data. The query used to get the number
        of rows is now changed to count only those rows that have
        no NULL values in the array.
        The mean of the dependent variable was still computed using
        an SQL command which was not ignoring the independent variables
        (arrays) with NULLs. They are ignored which computing the mean now.

@mktal

### DIFF
--- a/src/modules/convex/utils_regularization.cpp
+++ b/src/modules/convex/utils_regularization.cpp
@@ -79,13 +79,13 @@ class ScalesState
 AnyType utils_var_scales_transition::run (AnyType& args)
 {
     ScalesState<MutableArrayHandle<double> > state = args[0];
+    MappedColumnVector x(NULL);
     try {
-        args[1].getAs<MappedColumnVector>();
+        new (&x) MappedColumnVector(args[1].getAs<MappedColumnVector>());
     } catch(const ArrayWithNullException &e) {
       //  warning("Input array contains NULL values, skipping it for mean and std computations");
         return state;
     }
-    MappedColumnVector x = args[1].getAs<MappedColumnVector>();
     if (state.numRows == 0)
     {
         int dimension = args[2].getAs<int>();
@@ -157,13 +157,13 @@ AnyType __utils_var_scales_result::run (AnyType& args)
 
 AnyType utils_normalize_data::run (AnyType& args)
 {
+    MutableMappedColumnVector x(NULL);
     try {
-        args[0].getAs<MutableMappedColumnVector>();
+        new (&x) MutableMappedColumnVector(args[0].getAs<MutableMappedColumnVector>());
     } catch(const ArrayWithNullException &e) {
        // warning("Input array contains NULL values, skipping this array normalization");
         return Null();
     }
-    MutableMappedColumnVector x = args[0].getAs<MutableMappedColumnVector>();
     MappedColumnVector mean = args[1].getAs<MappedColumnVector>();
     MappedColumnVector std = args[2].getAs<MappedColumnVector>();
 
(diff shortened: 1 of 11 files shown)


src/modules/convex/utils_regularization.cpp
```diff
@@ -79,8 +79,13 @@ class ScalesState
 AnyType utils_var_scales_transition::run (AnyType& args)
 {
     ScalesState<MutableArrayHandle<double> > state = args[0];
+    try {
+        args[1].getAs<MappedColumnVector>();
+    } catch(const ArrayWithNullException &e) {
+      //  warning("Input array contains NULL values, skipping it for mean and std computations");
+        return state;
+    }
```
mktal: A more efficient way is to declare an empty handle outside the try block and rebind the memory later:

```c++
MappedColumnVector x(NULL);
try {
    new (&x) MappedColumnVector(args[1].getAs<MappedColumnVector>());
    // code ...
}
```

Save goes for other places where this applies.

src/ports/postgres/modules/convex/utils_regularization.py_in
```diff
@@ -47,14 +46,22 @@ def __utils_dep_var_scale(**kwargs):
...
+    col_ind_var -- independent variables column
+    col_dep_var -- dependent variable column
     """
+
     y_scale = plpy.execute(
         """
         select
-            avg({col_dep_var}) as mean,
+            avg(case when not {schema_madlib}.array_contains_null({col_ind_var}) then {col_dep_var} end) as mean,
             1 as std
```
mktal: What if `col_dep_var` contains `null`?

src/modules/convex/utils_regularization.cpp
njayaram2: Thank you, will change them.

src/ports/postgres/modules/convex/utils_regularization.py_in
njayaram2: I had tried inserting a row into the input table that contains a NULL for col_dep_var, and the insert had failed. So I assumed that kind of data is not allowed, and thus the check would not be necessary.
Should we still have a condition out there for col_dep_var?

decibel: That's not very defensive programming (assuming that the table column will always be NOT NULL).

BUT... I don't think that's necessary anyway. avg() should correctly handle NULLs:

```SQL
SELECT avg(unnest) FROM unnest('{1,2,NULL,NULL,NULL}'::int[]);
        avg         
--------------------
 1.5000000000000000
(1 row)
```

njayaram2: Thank you for the feedback!
But since avg() handles NULLs correctly as you pointed out, I will keep this as is.

mktal: @decibel that is a good point. I agree that for `avg()` function `col_dep_var` containing NULLs is not a problem. But what I meant is that it will be a problem for normalizing the `col_ind_var` because in that case we cannot treat `col_ind_var` and `col_dep_var` independently as we are currently doing.

PR #40, opened 2016-04-09
Fix to disable the redirect URL for downloading files.

snaga:
The redirect URL is no longer available.


fmcquillan99:
Thanks for catching this snaga!   We did a DNS change on Friday that broke the make.  But we have re-verted that change so the make should work OK now.  We will review and test your pull request and include in the 1.9.1 release to remove the unnecessary madlib.net hop.

Frank

iyerr3:
Merged with 6a5f177

PR #43, opened 2016-05-23
SVM: Add class weights for use with unbalanced data

iyerr3:
JIRA: MADLIB-998

Added 'class_weight' in the 'params' argument. It can either be a string or a
dictionary-like mapping. In case of a string, we currently only accept
'balanced' as an option. For a mapping, the user can map values of the
dependent variable to specific double precision weights. Since, SVM only
supports binary classification at present, the class_weight mapping can only
take upto two arguments.

As part of this work, we add a 'tuple_weight' argument to the SVM aggregate.
This allows future addition of sample weights (which would be multiplied with
the class weight).


iyerr3:
Testing with scikit learn revealed odd behavior. See figures below.
1) With equal class size the results are similar between scikit and madlib.
2) With unequal class sizes and class_weight = balanced the results are also close enough (dashed line). But the hyperplane without class weights are quite different with MADlib being obviously wrong.

with_weights
no_weights

iyerr3:
Things are better when we don't use the default step size. 

Using init_stepsize=0.1 instead of the default (0.01): 

with_weights_higher_init

We need to investigate the optimal learning schedule used in scikit-learn, where the initial stepsize is determined based on a heuristic proposed by Léon Bottou.

fmcquillan99:
Thanks for posting the the pics.  I would suggest we create another JIRA to look at the heuristic proposed by Léon Bottou.  I'll do that.

src/ports/postgres/modules/svm/svm.py_in
```diff
@@ -781,6 +785,46 @@ def _random_feature_map(schema_madlib, source_table, dependent_varname,
...
+        # use half of n_samples since only doing binary classification
+        # Change the '2' to n_classes for multinomial
+        n_samples_per_class = num_samples(source_table) / 2
+        bin_count = plpy.execute("""SELECT {dep} as k, count(*) as v
+                                    FROM {src}
+                                    GROUP BY {dep}
+                                 """.format(dep=dependent_varname,
+                                            src=source_table))
+        for each_count in bin_count:
+            plpy.info(each_count)
```
iyerr3: remove the 'info' line

PR #44, opened 2016-05-31
Feature: Sessionize funtion

njayaram2:
JIRA: MADLIB-909

This contains the implementation of the phase 1 sessionize function.
The current input parameters are the input and output table names,
along with a partition expression, the time_stamp column name and
the time_out period to consider for sessionization. The implementation
uses the window function to perform sessionization.
This commit also contains the install check file for sessionization.

### DIFF
--- a/src/ports/postgres/modules/utilities/sessionize.py_in
+++ b/src/ports/postgres/modules/utilities/sessionize.py_in
@@ -1,0 +1,101 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import plpy
+import string
+
+from control import MinWarning
+from utilities import unique_string, _assert
+from validate_args import get_cols
+from validate_args import input_tbl_valid, output_tbl_valid, is_var_valid
+
+m4_changequote(`<!', `!>')
+
+def sessionize(schema_madlib, source_table, output_table, partition_expr,
+				time_stamp, time_out, **kwargs):
+	"""
+		Perform sessionization over a sequence of rows.
+
+		Args:
+		@param schema_madlib: str, Name of the MADlib schema
+		@param source_table: str, Name of the input table/view
+		@param output_table: str, Name of the table to store result
+		@param partition_expr: str, Expression to partition (group) the input data
+		@param time_stamp: float, Column name with time used for sessionization calculation
+		@param time_out: float, Delta time between subsequent events to define a sessions
+		
+	"""
+	with MinWarning("error"):
+		_validate(source_table, output_table, partition_expr, time_stamp, time_out)
+
+		all_input_cols_str = ', '.join([i.strip() for i in get_cols(source_table, schema_madlib)])
+		session_id = 'session_id' if not is_var_valid(source_table, 'session_id') else unique_string('session_id')
+
+		plpy.execute("""
+				CREATE TABLE {output_table} AS
+					SELECT
+						{all_input_cols_str},
+						CASE WHEN {time_stamp} NOTNULL
+						THEN (SUM(new_event_boundary) OVER (PARTITION BY {partition_expr} ORDER BY {time_stamp})) END AS {session_id}
+					FROM (
+						SELECT *, 
+							CASE WHEN {time_stamp} NOTNULL and ({time_stamp}-LAG({time_stamp},1) OVER (w) > '{time_out}' OR ROW_NUMBER() OVER (w) = '1')
+							THEN 1 ELSE 0 END AS new_event_boundary
+						FROM {source_table} WINDOW w AS (PARTITION BY {partition_expr} ORDER BY {time_stamp})
+						) a
+			""".format(**locals()))
+
+
+def _validate(source_table, output_table, partition_expr, time_stamp, time_out):
+	input_tbl_valid(source_table, 'Sessionization')
+	output_tbl_valid(output_table, 'Sessionization')
+	# ensure the expressions are not None or empty strings
+	_assert(partition_expr, "Sessionization error: Invalid partition expression")
+	_assert(time_stamp, "Sessionization error: Invalid time stamp column")
+	_assert(time_out, "Sessionization error: Invalid time out value")
+	# ensure the partition/order expression can actually be used
+	_assert(is_var_valid(source_table, partition_expr, time_stamp),
+			"Sessionization error: invalid partition expression or time stamp column name")
+
+
+def sessionize_help_message(schema_madlib, message, **kwargs):
+	"""
+	Help message for sessionize function
+	"""
+	help_string = """
+------------------------------------------------------------
+						SUMMARY
+------------------------------------------------------------
+Functionality: Sessionize
+
+The goal of the MADlib sessionize function is to perform sessionization over
+a time-series based data. 
+
+------------------------------------------------------------
+						USAGE
+------------------------------------------------------------
+SELECT {schema_madlib}.sessionize(
+	'source_table',    -- Name of the table
+	'output_table',    -- Table name to store the Sessionization results
+	'partition_expr',  -- Partition expression to group the data table
+	'time_stamp'	-- Column name with time used for sessionization calculation
+	'time_out'	-- Delta time between subsequent events to define a sessions
+				-- User should make sure in same units the min_time parameter (e.g., min, sec, days)
+);
+"""
+
+	return help_string.format(schema_madlib=schema_madlib)
--- a/src/ports/postgres/modules/utilities/sessionize.py_in
+++ b/src/ports/postgres/modules/utilities/sessionize.py_in
@@ -26,7 +26,7 @@ from validate_args import input_tbl_valid, output_tbl_valid, is_var_valid
 m4_changequote(`<!', `!>')
 
 def sessionize(schema_madlib, source_table, output_table, partition_expr,
-				time_stamp, time_out, **kwargs):
+				time_stamp, max_time, **kwargs):
 	"""
 		Perform sessionization over a sequence of rows.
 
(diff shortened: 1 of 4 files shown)


decibel:
Has consideration been given to allowing for the creation of a view instead of a table? If you only needed session info once, or only for a subset of values in the input table, that could be significantly faster than materializing it.

Related to that... it could also be faster to only materialize the partition columns, series number and timespan for each series. You would need to join the base data against that, but in some cases it could be faster. A variation on this would be to specify the exact columns you want materialized.

So perhaps add "create_view = False, materialize_columns = None" parameters, where materialize_columns == None means materialize the whole shebang, while materialize_columns = [] means only materialize the partition clause, session and timestamp.

njayaram2:
@decibel Thank you for the comment. This is certainly useful, and I am planning to implement this as part of sessionization-Phase 2 (I will put in the link of the corresponding JIRA soon). The following are the changes I plan to make based on your comment:
1) Have an optional "params" parameter which would include:
- **create_view** (T/F): default is **true**
   - (TRUE) will create a view, instead of table.
   - (FALSE) will create a table. The name for the view/table will be specified in the output_table parameter.

- **output_all_columns** (T/F): default is **fault**.
   - (FALSE) will output the partition columns, time_stamp column and a new session ID column. The assumption is that the partition columns with the time_stamp column will be sufficient to perform a join with the input table.
   - (TRUE) will output all the columns in the table, along with the new session_id column.

2) Other optional parameters such as **min_time** and **persist_null** can also be specified in "params" (details of these will be updated in the JIRA). 

Please let me know if you have any other comments.

decibel:
I'm not sure why params is better than explicit parameters, but lets move that discussion to the new Jira.

I didn't do a full review, but the database stuff looks sane.

fmcquillan99:
Here are the follow on JIRAs on sessionization from product mgmt perspective.  Others can chip in on performance and implementation issues, and open other JIRAs as required.

Sessionization - Phase 2 (output controls)
https://issues.apache.org/jira/browse/MADLIB-1001

Sessionization - Phase 3 (minimum time)
https://issues.apache.org/jira/browse/MADLIB-1002

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,101 @@
...
+		_validate(source_table, output_table, partition_expr, time_stamp, time_out)
+
+		all_input_cols_str = ', '.join([i.strip() for i in get_cols(source_table, schema_madlib)])
+		session_id = 'session_id' if not is_var_valid(source_table, 'session_id') else unique_string('session_id')
+
+		plpy.execute("""
+				CREATE TABLE {output_table} AS
+					SELECT
+						{all_input_cols_str},
+						CASE WHEN {time_stamp} NOTNULL
```
decibel: Is it necessary to include rows with NULL timestamps? While that's not a big deal here, I think it's going to lead to unwanted complexity down the road (I'm looking at JIRA MADLIB-1002).

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,101 @@
...
+						THEN (SUM(new_event_boundary) OVER (PARTITION BY {partition_expr} ORDER BY {time_stamp})) END AS {session_id}
+					FROM (
+						SELECT *, 
+							CASE WHEN {time_stamp} NOTNULL and ({time_stamp}-LAG({time_stamp},1) OVER (w) > '{time_out}' OR ROW_NUMBER() OVER (w) = '1')
```
decibel: Using '1' is bad here, because it will force a cast for every comparison, and it results in a potentially extremely expensive conversion-aware string comparison.

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,101 @@
...
+				time_stamp, time_out, **kwargs):
+	"""
+		Perform sessionization over a sequence of rows.
+
+		Args:
+		@param schema_madlib: str, Name of the MADlib schema
+		@param source_table: str, Name of the input table/view
+		@param output_table: str, Name of the table to store result
+		@param partition_expr: str, Expression to partition (group) the input data
+		@param time_stamp: float, Column name with time used for sessionization calculation
```
decibel: I'm assuming this is not actually a float...

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,101 @@
...
+		@param time_out: float, Delta time between subsequent events to define a sessions
```
decibel: time_out makes me think this is "time out", not "timeout". If timeout is the desired term then the underscore should go.

Another possibility would be "max_delta" or "max_time_delta". That would provide nice symmetry with JIRA: MADLIB-1002.

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,101 @@
...
+
+		all_input_cols_str = ', '.join([i.strip() for i in get_cols(source_table, schema_madlib)])
+		session_id = 'session_id' if not is_var_valid(source_table, 'session_id') else unique_string('session_id')
+
+		plpy.execute("""
+				CREATE TABLE {output_table} AS
+					SELECT
+						{all_input_cols_str},
+						CASE WHEN {time_stamp} NOTNULL
+						THEN (SUM(new_event_boundary) OVER (PARTITION BY {partition_expr} ORDER BY {time_stamp})) END AS {session_id}
```
decibel: I did some refactoring of this while taking a look at https://issues.apache.org/jira/browse/MADLIB-1002. Please take a look at it, as I think it's clearer than this code. *Note that I have not tested it for correctness!*

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: You are right, that's a typo!

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: @decibel great comments! I think the refactored code you have put up in MADLIB-1002 incorporates most of the comments you have made here. I will also include the type casting suggestion and make the changes.

If the min_time solution you have proposed in https://issues.apache.org/jira/browse/MADLIB-1002 works fine, do you think we should merge the sessionization phases 1 and 3?

decibel: > If the min_time solution you have proposed in
> https://issues.apache.org/jira/browse/MADLIB-1002 works fine, do you
> think we should merge the sessionization phases 1 and 3?

Seems logical to me, but I'm not the one doing the work... ;)

njayaram2: I quickly tried the query with min_time proposed in MADLIB-1002, and it seems to be missing one of the requirements. Guess phase 3 might indeed have to stay as planned! :-/

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: Are you suggesting we mandate that the input table has no NULL timestamps?

I tried leaving out the NOT NULL mentioned in the comment, but it will end up showing a session id for rows with NULL timestamps too if I do that. I wanted to make sure rows with NULL timestamps have no session ID at all, so had included the NOT NULL along with the NOT NULL inside the subselect.

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: _Likewise, it will be a lot more efficient if '{time_out}' is cast to the correct data type._

Consider the example where {time_stamp} = "'04/15/2015 01:03:0.5'" and LAG({time_stamp},1) = "'04/15/2015 01:05:00'". If the {time_out} = "0:3:0", the following operation tries to check if the difference between the two time_stamps is greater than 3 minutes or not:
({time_stamp}-LAG({time_stamp},1) OVER (w) > '{time_out}'

Trying to cast {time_out} to timestamp (in this example) throws an error. I cannot expect {time_out} to be of type timestamp, since the difference between two timestamps seems to be of type interval. If time_stamp was of type date instead, having a time_out/min_time of type date may have similar issues.

Enforcing the type of time_out/min_time to "interval" (instead of varchar or the type of time_stamp) might be a better approach, since time_stamp can then be any of the date/time types.

Do you think we should mandate time_out/min_time to always be of type interval? Another option is to force it to be of type float (representing seconds), then we may have to use epoch() while computing the time_stamp diffs.

decibel: > Do you think we should mandate time_out/min_time to always be of type
> interval? Another option is to force it to be of type float
> (representing seconds), then we may have to use epoch() while computing
> the time_stamp diffs.

On the Postgres side, interval is what's wanted.

On the python side, I could see use for either of them. float might be 
the simplest, but there are options available with interval that you 
don't have with seconds, because interval tracks seconds, days, and 
months separately. In this specific case I can't see days or months 
being very useful, but from an overall MADlib API standpoint I think 
it's best if anything that accepts a time delta can accept an interval.

Code-wise, I think the best way to handle this is to see if the python 
interval parameter is some kind of number and if it is pass it directly 
to Postgres and multiply it by a 1 second interval, ie: "{time_out} * 
interval '1 second'". Postgres will do the correct thing regardless of 
what numeric type it is (int vs numeric vs float). Otherwise just assume 
it's something that can be cast to an interval and do 
"{time_out}::interval".

njayaram2: _Code-wise, I think the best way to handle this is to see if the python
interval parameter is some kind of number and if it is pass it directly
to Postgres and multiply it by a 1 second interval, ie: "{time_out} *
interval '1 second'". Postgres will do the correct thing regardless of
what numeric type it is (int vs numeric vs float). Otherwise just assume
it's something that can be cast to an interval and do
"{time_out}::interval"._

If I understood you correctly, {time_out} and {min_time} would still be strings. They can 
either have a value such as '30', in which case we would internally convert it to 30 seconds, 
or it can have a string with non-numeric characters which will be cast to type interval 
(the casting will fail if the string is not in line with interval's format).

But, Postgres' interval type seems to consider something like '30' to be a legitimate interval
value (I reckon it is considered as 30 hours). Will have to check if converting '30' to seconds
instead of letting {time_out}::interval handle it is acceptable from the API's perspective.

njayaram2: I just tried out a couple of more queries, and it looks like '30' is automatically converted
to 30 seconds by {time_stamp}::interval casting (not 30 hours as I previously stated). 
So we don't have to do {time_out}* interval '1 second' for a {time_out} which 
only has some kind of number in it.

I think we can just mandate the {time_out} and {min_time} parameters to be of type
interval, and of course, cast it to interval in the query.

On a side note, '30:0' is converted to 30 hours when we cast it to interval.

decibel: > I think we can just mandate the {time_out} and {min_time} parameters to
> be of type
> interval, and of course, cast it to interval in the query.

Just to clarify my original intent, I was thinking of something like:

try:
   float(time_out)
   time_out_is_number = True
except ...:
   time_out_is_number = False

But I agree that just passing it directly to Postgres and letting it 
cast it to an interval is the simplest and best approach.

njayaram2:
I have made the changes based on the various discussions we've had over the comments.
The changes made are:
- time_out is renamed to max_time
- max_time is of type INTERVAL and not VARCHAR anymore
- the window function query itself is changed to a cleaner version
as suggested by @decibel

iyerr3:
Use 4 spaces instead of 1 tab for indentation. I would suggest setting this as default in editor. More info in the [Python style guide](https://cwiki.apache.org/confluence/display/MADLIB/Python+Style+Guide)

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -0,0 +1,103 @@
...
+------------------------------------------------------------
+SELECT {schema_madlib}.sessionize(
+	'source_table',    -- str, Name of the table
+	'output_table',    -- str, Table name to store the Sessionization results
+	'partition_expr',  -- str, Partition expression to group the data table
+	'time_stamp'	-- str, Column name with time used for sessionization calculation
+	'max_time'	-- str, Delta time between subsequent events to define a session
+);
+"""
+
```
iyerr3: it would be good to have 1 of the examples from docs here as well.

orhankislal:
The documentation does not create a new page for Sessionize. It just adds the text to the end of Path docs. It also isn't formatted properly (especially the results). Here are two screenshots demonstrating the issues. 

screen shot 2016-06-08 at 3 59 24 pm
screen shot 2016-06-08 at 3 58 52 pm

PR #49, opened 2016-06-20
Feature: Sessionize funtion - Phase 2

njayaram2:
JIRA: MADLIB-1001

This contains the phase 2 implementation of the sessionize function.
This commit adds two new optional parameters: output_cols and
create_view. output_cols is a comma separated list of columns to
be projected into the output, while create_view indicates if the
output is a view or a table. The default values of output_cols
and create_view are '*' and True respectively.

These two parameters are for performance reasons. A view is
generally faster when compared to materializing the results to an
actual table, while having to include all columns when not necessary
can be avoided with the output_cols parameter.

### DIFF
--- a/src/ports/postgres/modules/utilities/sessionize.py_in
+++ b/src/ports/postgres/modules/utilities/sessionize.py_in
@@ -50,7 +50,7 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
         table_or_view = 'VIEW' if create_view or create_view is None else 'TABLE'
         output_cols_to_project = '*' if output_cols is None else output_cols
 
-        cols_to_project = get_column_names(schema_madlib, source_table, output_cols_to_project)
+        cols_to_project = get_columns_to_project_str(schema_madlib, source_table, output_cols_to_project)
         session_id = 'session_id' if not is_var_valid(source_table, 'session_id') else unique_string('session_id')
 
         # Create temp column names for intermediate columns.
@@ -71,9 +71,22 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
                         ) a
             """.format(**locals()))
 
-def get_column_names(schema_madlib, source_table, output_cols):
+def get_columns_to_project_str(schema_madlib, source_table, output_cols):
     """
-        This method creates a string that can be used in the SQL statement to project the columns specified in the output_cols parameter.
+        This method creates a string that can be used in the SQL statement to project the columns 
+        specified in the output_cols parameter.
+        Args:
+        @param schema_madlib: str, Name of the MADlib schema
+        @param source_table: str, Name of the input table/view
+        @param output_cols: str, comma separated string specifying coulumns/expressions to project
+
+        If an element in output_cols is a column name from the input table, it is maintained as is.
+        But, if it is an expression, the expression is renamed using a quoted string.
+
+        Example:
+        output_cols = 'user_name, "user id"<100", revenue>20'
+        Return value:
+        'user_name, "user id"<100 AS "user id<100", revenue>20 AS "revenue>20"'
 
         Return:
         a string to be used in the SQL statement
@@ -89,12 +102,36 @@ def get_column_names(schema_madlib, source_table, output_cols):
     return output_cols_str
 
 def create_column_name_from_expression(col_name, table_columns_list):
+    """
+    This method creates a new name for an expression, which is used as the column name for that expression
+    when materialized into the result table/view.
+
+    Args:
+    @param: col_name: str, the expression.
+    @param: table_columns_list: list, contains all the existing columns names of the input table
+
+    All existing quotes in col_name, if any, are removed. The string thus obtained is then quoted. If this quoted
+    string is already an element in table_columns_list, then a unique string is created as the column name corresopnding
+    to col_name.
+
+    Example:
+    1) col_name= "user id"<100
+    Return value: "user id<100"
+    2) col_name = revenue>20
+    Return value: "revenue>20"
+
+    Return:
+    A string which represents the column name corresponding to the expression specified in col_name
+    """
     if col_name == '*':
         return get_cols_str(table_columns_list)
     else:
-        # Column name cannot have more than one pair of quotes in it. Removing any existing quotes, and then quoting the new string obtained.
+        # Column name cannot have more than one pair of quotes in it. Removing any existing quotes,
+        # and then quoting the new string obtained.
         col_name = col_name.replace('"','')
         col_name = '"'+col_name + '"'
+        # If the col_name thus obtained is the same as an existing column name in the input table,
+        # return a unique string, else return col_name
         if col_name in table_columns_list:
             return unique_string(col_name)
         else:
@@ -104,7 +141,13 @@ def get_cols_str(table_columns_list):
     return ', '.join([i for i in table_columns_list])
 
 def get_columns_from_expression(output_cols, table_columns_list):
+    """
+    output_cols is a comma separated string of expressions. This function creates a list consisting of the elements (expressions)
+    in output_cols, and another list containing the corresponding name for each element (expression).
+    """
     cols_list = split_quoted_delimited_str(output_cols)
+    # If an entry in cols_list is an existing column name in the input table, keep the name as is. If NOT,
+    # create a new name for the entry (expression).
     cols_names = [i if i in table_columns_list else create_column_name_from_expression(i, table_columns_list) for i in cols_list]
     return cols_list, cols_names
 
--- a/src/ports/postgres/modules/utilities/sessionize.py_in
+++ b/src/ports/postgres/modules/utilities/sessionize.py_in
@@ -46,20 +46,29 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
     with MinWarning("error"):
         _validate(source_table, output_table, partition_expr, time_stamp, max_time)
         table_or_view = 'VIEW' if create_view or create_view is None else 'TABLE'
-
         output_cols = '*' if output_cols is None else output_cols
-        # If the output_cols has '*' as one of the elements, expand it to include all columns in the source table. The following list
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -35,41 +36,83 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
...
+        _validate_output_cols(source_table, output_cols_list)
+        output_cols_str = ', '.join([output_cols_names[i] if output_cols_list[i] == '*' else output_cols_list[i] + ' AS ' + output_cols_names[i] 
+            for i in range(len(output_cols_list))])
+    return output_cols_str
+
+def create_column_name_from_expression(col_name, table_columns_list):
+    if col_name == '*':
+        return get_cols_str(table_columns_list)
+    else:
+        # Column name cannot have more than one pair of quotes in it. Removing any existing quotes, and then quoting the new string obtained.
```
decibel: This isn't going to work. One immediate test case that comes to mind is "column with spaces in the name".

If you wanted to do something around quoting, I think you need to do everything that Postgres quote_ident() does.

How is this handled elsewhere in MADlib? What happens if you have "a table with spaces in the name"?

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -35,41 +36,83 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
...
+    if output_cols.strip() == '*':
+        output_cols_str = get_cols_str(table_columns_list)
+    else:
+        output_cols_list, output_cols_names = get_columns_from_expression(output_cols, table_columns_list)
+        _validate_output_cols(source_table, output_cols_list)
+        output_cols_str = ', '.join([output_cols_names[i] if output_cols_list[i] == '*' else output_cols_list[i] + ' AS ' + output_cols_names[i] 
+            for i in range(len(output_cols_list))])
+    return output_cols_str
+
+def create_column_name_from_expression(col_name, table_columns_list):
```
decibel: Since this returns a list of columns, I think it would be better to call it something like column_names_from_expression

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -35,41 +36,83 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
         @param source_table: str, Name of the input table/view
         @param output_table: str, Name of the table to store result
         @param partition_expr: str, Expression to partition (group) the input data
-        @param time_stamp: str, Column name with time used for sessionization calculation
+        @param time_stamp: str, The time stamp column name that is used for sessionization calculation
         @param max_time: interval, Delta time between subsequent events to define a session
-
+        @param output_cols: str, list of columns the output table/view must contain (default '*'):
+                        * - all columns in the input table, and a new session ID column
+                        'a,b,c,...' -  a comma separated list of column names/expressions to be projected, along with a new session ID column
```
decibel: Would it be better to accept a generator if someone wanted to list column names? That certainly seems more pythonic than a string...

src/ports/postgres/modules/utilities/test/sessionize.sql_in
```diff
@@ -87,6 +89,23 @@ SELECT
...
+FROM sessionize_output_v;
+SELECT * FROM sessionize_output_v;
+
+SELECT sessionize(
+        'eventlog_installchk', -- Name of the input table
+        'sessionize_output_t', -- Name of the output table
+        'user_id<102000', -- Partition expression to group the data
+        'event_timestamp', -- Order expression to sort the tuples of the data table
+        '180', -- Max time that can elapse between consecutive rows to be considered part of the same session
+        '*,user_id<102000',  -- Select all columns in the input table, along with the partition expression and session id columns
```
decibel: Supporting arbitrary expressions in the column list is going to lead to confusion (or worse), because Postgres will call that column something else. I'm pretty sure this will break if you tried to turn it into a table, and I know it'll break if you try adding 2 expressions.

Instead of trying to treat this as a column list, maybe it would be better to simply treat it as a replacement for the auto-generated select clause and be done with it. Does anything in code actually depend on having the list of columns?

Something to ponder... a way to "cheat" and turn an arbitrary select clause into the columns that it will actually result in is to create a temporary view and then inspect the catalog to see what columns you ended up with. You can then drop the temp view.

src/ports/postgres/modules/utilities/sessionize.py_in
decibel: While looking at #47, I noticed that there are already column parsing utilities. See https://github.com/apache/incubator-madlib/pull/47/files#diff-96a0a5136d71ec812cc09b1d3a2e259aR35.

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: Will rename this and other functions too. It seems to be causing a lot of confusion with existing
utility functions used for parsing.

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: I am not sure I understand the comment exactly. This method does take care of column names with spaces in the name. Say I had a column name called "user id" (note that it is quoted, since postgres does not let me name a column with spaces without quotes), and another column called revenue. 
If my output_cols parameter was:
' "user id"<100, revenue>20 '

I parse this to rename each expression:
- "user id"<100 is named as "user id<100"
- revenue>20 is named as "revenue>20"

This ends up creating a select statement as follows:
SELECT "user id"<100 AS "user id<100", revenue>20 AS "revenue>20"

All I am doing in this function is creating a new column name for an expression. I quote the expression to create a new column name, but before doing that, I remove any existing quotes from the expression (because postgres does not let me have quotes inside a quoted column name). 

I checked again to make sure this is indeed how its working. I will put up some install checks to cover these cases too.

src/ports/postgres/modules/utilities/test/sessionize.sql_in
njayaram2: I rename each expression using quotes, as mentioned in one of my comments above. Since I rename each expression and don't rely on postgres giving a default name for an expression, the code will handle cases where there are 2 or more expressions. But I will put up better comments and test cases in install check.
Reproducing the example from the previous comment here for context:
If my output_cols parameter was:
' "user id"<100, revenue>20 '

I parse this to rename each expression:

"user id"<100 is named as "user id<100"
revenue>20 is named as "revenue>20"
This ends up creating a select statement as follows:
SELECT "user id"<100 AS "user id<100", revenue>20 AS "revenue>20"

src/ports/postgres/modules/utilities/sessionize.py_in
njayaram2: I believe we had precedent in other MADlib functions, and hence chose to go with the comma separated string to specify output_cols. 
One potential modification could be to ask the user to specify a valid SELECT expression as output_cols, with the constraint that expressions must be renamed using AS. For instance, output_cols should be something like the following:
'*, "user id"<100 AS uid_100, revenue>20 AS rev_20',
instead of its current form which is: 
'*, "user id"<100, revenue>20'

We will have to decide if this is too hard a constraint to have or not. Having this constraint will take away all the messy string parsing stuff we currently have implemented though.

src/ports/postgres/modules/utilities/sessionize.py_in
decibel: >   * "user id"<100 is named as "user id<100"
>   * revenue>20 is named as "revenue>20"

Ahh, ok. That was completely unclear to me. It would be good to add 
comments/docstrings to the various functions explaining what they're doing.

iyerr3: +1. 
I suggest also adding the above examples to the docstrings.

src/ports/postgres/modules/utilities/sessionize.py_in
decibel: > We will have to decide if this is too hard a constraint to have or not.
> Having this constraint will take away all the messy string parsing stuff
> we currently have implemented though.

Yeah, I didn't realize what the code was doing. I think it would be nice 
to allow the user to choose different output names if they want.

Perhaps a good compromise would be to detect ' AS ' in the string and 
then treat it as a raw select clause. Another option would be to treat 
an array as a list of columns and anything else as a select clause 
(which would also support the * case, I think).

njayaram2:
Thank you for the comments @decibel  and @iyerr3. I will improve the install check and documentation,
and update the pull request.

fmcquillan99:
Thanks for the discussion on this pull request.  I would suggest that asking the user to provie a valid SELECT expression as output_cols is not unreasonable.  We just need to be clear in the docs.

For example, in the path function http://madlib.incubator.apache.org/docs/latest/group__grp__path.html#examples
we do a similar thing for the partition parameter.

njayaram2:
I have pushed the latest code that treats output_cols as a valid select expression. I have removed
the code that was doing a lot of string parsing to rename expressions, and the onus is now on the
user to use ' AS ' and rename expressions in output_cols.

src/ports/postgres/modules/utilities/sessionize.py_in
```diff
@@ -35,41 +36,57 @@ def sessionize(schema_madlib, source_table, output_table, partition_expr,
...
+        @param create_view: boolean, indicates if the output is a view or a table with name specified by output_table (default TRUE)
+                        TRUE - create view
+                        FALSE - materialize results into a table
     """
     with MinWarning("error"):
         _validate(source_table, output_table, partition_expr, time_stamp, max_time)
+        table_or_view = 'VIEW' if create_view or create_view is None else 'TABLE'
+
+        output_cols = '*' if output_cols is None else output_cols
+        # If the output_cols has '*' as one of the elements, expand it to include all columns in the source table. The following list
```
iyerr3: This block needs to be wrapped into 72 characters and appropriate line space needs to be added to improve readability.

PR #45, opened 2016-06-13
Path: refactor with dynamic link to boost regex

mktal:

PR #51, opened 2016-06-28
Path: Handling multiple expressions in partition_expr

njayaram2:
JIRA: MADLIB-1003

The various expressions in the partition_expr parameter are now
quoted and renamed. These expressions were not renamed earlier,
which led to an error when there were multiple expressions in the
parameter. Renaming them using ' AS ' fixes this issue.
This commit also includes an update to the install check, which
now includes multiple expressions in partition_expr.

### DIFF
--- a/src/ports/postgres/modules/utilities/path.py_in
+++ b/src/ports/postgres/modules/utilities/path.py_in
@@ -37,7 +37,6 @@ from validate_args import quote_ident
 
 m4_changequote(`<!', `!>')
 
-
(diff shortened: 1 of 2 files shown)


iyerr3:
LGTM but I'm sure there will be conflicts between this and #45. I suggest you rebase on top of that and make any necessary changes before we merge this story (or wait till #45 is merged and then rebase).

njayaram2:
I think I will wait for #45 to be merged and then rebase it with that.

PR #54, opened 2016-07-06
Pivoting: Phase 2

orhankislal:
JIRA: MADLIB-1004

This phase of the pivoting development adds the following functionality

- Multiple indices
- Multiple pivot columns
- Multiple value columns
- Multiple aggregate functions
- Value column specific aggregate functions
- Keep null values in pivot columns
- Fill null values in the output table

### DIFF
--- a/src/ports/postgres/modules/utilities/pivot.py_in
+++ b/src/ports/postgres/modules/utilities/pivot.py_in
@@ -30,7 +30,6 @@
 @file pivot.py_in
 
 """
-
 import plpy
 import itertools
 from control import MinWarning
@@ -98,36 +97,19 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
     pcols = [strip_end_quotes(pcol.strip()) for pcol in pcols]
     pvals = [strip_end_quotes(pval.strip()) for pval in pvals]
 
-    # Parse the aggregate_func as a dictionary
-    try:
-        param_types = dict.fromkeys(pvals, list)
-        agg_dict = extract_keyvalue_params(aggregate_func,param_types)
-    except KeyError, e:
-        with MinWarning("warning"):
-            plpy.warning("Pivot: Not all columns from '{aggregate_func}'"
-            " present in '{pivot_values}'".format(**locals()))
-            raise
-
-    # If the dictionary is empty, parse it as a list
-    agg_set = split_quoted_delimited_str(aggregate_func) if not agg_dict else []
-
-    # __madlib_pivot_def_agg__ denotes the aggregate function(s) if the user
-    # does not specify a value: aggregate dictionary
-    # If no aggregates are given, set average as default
-    agg_dict['__madlib_pivot_def_agg__'] = ['avg'] if not agg_set else agg_set
+    # Create a dictionary that assigns one or more aggregate functions for every
+    # value column.
+    agg_dict = parse_aggregates(pvals, aggregate_func)
 
     # Find the distinct values of pivot_cols
-    # Note that the distinct values are appended in order of users list
-    # This ordering is important when we access pivot_comb entries
     array_agg_str = ', '.join("array_agg(DISTINCT {pcol}) AS {pcol}_values".
         format(pcol=pcol) for pcol in pcols)
-
     null_str = ""
     if keep_null:
         # Create an additional column for every pivot column
         # If there is a null value, this column will get True
         null_str = ","+', '.join(
-            "bool_or(CASE WHEN {pcol} IS NULL THEN True END)"
+            "bool_or(CASE WHEN {pcol} IS NULL THEN True ELSE False END)"
             "AS {pcol}_isnull".format(pcol=pcol) for pcol in pcols)
 
     distinct_values = plpy.execute("SELECT {0} {1} FROM {2}".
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,164 @@ def pivot(schema_madlib, source_table, out_table,
...
+
+    # Collect the distinc values for every pivot column
+    pcol_distinct_values = []
+    for pcol in pcols:
+        # Read the distinct values for this pcol
+        pcol_tmp = [item for item in
+            distinct_values[0][pcol+"_values"]]
+        # Remove null values if keep null is not true
+        if not keep_null:
+            pcol_tmp = filter(None,pcol_tmp)
```
orhankislal: This removes 0 values in addition to NULLs. I replaced it with `pcol_tmp = [x for x in pcol_tmp if x is not None]`

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
-    pcol_no_quotes = strip_end_quotes(pcol[0].strip())
-    pval_no_quotes = strip_end_quotes(pval[0].strip())
+    # Strip end quotes from pivot columns
+    pcols = [strip_end_quotes(pcol.strip()) for pcol in pcols]
+    pvals = [strip_end_quotes(pval.strip()) for pval in pvals]
+
+    # Parse the aggregate_func as a dictionary
+    try:
+        param_types = dict.fromkeys(pvals, list)
+        param_types['__madlib_def_pval__'] = list
```
iyerr3: why do we need this?

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
+        agg_dict = extract_keyvalue_params(aggregate_func,param_types)
+    except KeyError, e:
+        with MinWarning("warning"):
+            plpy.warning("Pivot: Not all columns from '{aggregate_func}'"
+            " present in '{pivot_values}'".format(**locals()))
+            raise
+
+    # If the dictionary is empty, parse it as a list
+    agg_set = split_quoted_delimited_str(aggregate_func) if len(agg_dict) < 1 \
```
iyerr3: `... if not agg_dict ...`

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
+         else []
+    # If the list is empty set 'avg' as default
+    agg_dict['__madlib_def_agg__'] = ['avg'] if len(agg_set) < 1 else agg_set
```
iyerr3: `... if not agg_set ...`
We need explanation on the special name.

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
-                  format(pcol=pcol_no_quotes, value=str(value))
-                  for value in distinct_values if value is not None))
+    # Note that the distinct values are appended in order of users list
+    # This ordering is important when we access pivot_comb entries
+    distinct_str = ["SELECT "]
+    distinct_str.append(
+        ', '.join("array_agg(DISTINCT {pcol}) AS {pcol}_values".
+            format(pcol=pcol) for pcol in pcols))
+    distinct_str.append(" FROM " + source_table)
+    distinct_values = plpy.execute(''.join(distinct_str))
```
iyerr3: Why not create a single query string instead of joining them? 
```
array_agg_str = ', '.join("array_agg(DISTINCT {pcol}) AS {pcol}_values".format(pcol=pcol) for pcol in pcols)
plpy.execute("SELECT {0} FROM {1}".format(array_agg_str, source_table))
```

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
+
+    # Collect the distinc values for every pivot column
```
iyerr3: `s/distinc/distinct`

src/ports/postgres/modules/utilities/pivot.py_in
orhankislal: It can be removed. I moved the default aggregate but forgot to remove this line.

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
+        pcol_distinct_values[pcol]=pcol_tmp
+        # Max length of the string that pcol values can create +
+        # length of pcol + 1 (for _ character)
+        pcol_max_length+=max([len(str(item)) for item in pcol_tmp])+len(pcol)+1
+
+    # For non postgres systems add null to the pcol_distinct_values if necessary
+    # This is required because DISTINCT keyword is implemented differently
+    is_pg = m4_ifdef(<!__POSTGRESQL__!>, True, False)
+    if keep_null and not is_pg:
+        null_str_begin = ["SELECT "]
```
iyerr3: Again, this would be cleaner with the whole query as a single string. 
Also you could simplify it by using `EXISTS (SELECT 1 FROM {tbl} WHERE {pcol} is NULL) as {p_col}_isnull, ...`

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,255 @@ def pivot(schema_madlib, source_table, out_table,
...
+        null_values = plpy.execute(null_str)
+        for pcol in pcols:
+            pcol_tmp = [item for item in null_values[0][pcol+"_isnull"]]
+            if 1 in pcol_tmp:
+                pcol_distinct_values[pcol].append(None)
+
+    # Create the combination of every possible pivot column
+    # Assume piv and piv2 are pivot columns. piv=(1,2) and piv2=(3,4,5)
+    # pivot_comb = ((1,3),(1,4),(1,5),(2,3),(2,4),(2,5))
+    pivot_comb = list(itertools.product(*([pcol_distinct_values[pcol]
```
iyerr3: I suggest keeping the product output as an iterator (i.e. don't make it a list) since you're only looping through it once. Further keep a count of number of pivot_comb seen while iterating and return a warning when you've reached the threshold. Finally, make the threshold a constant and initialize it at top with explanation on why it's set to that particular value (1000).

orhankislal: Thank you very much for the comments @iyerr3. The calculation for threshold happens if the code throws an exception. List is accessed multiple times if there are multiple value columns or aggregates. I fixed rest of the issues you pointed out.

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -58,66 +61,256 @@ def pivot(schema_madlib, source_table, out_table,
...
+
+    # Check the max possible length of a output column name
+    # If it is over 63 (psql upper limit) create table lookup
+    for pval in pvals:
+
+        col_name_len = pcol_max_length+len(pval)+1
+        try:
+            # If user specifies a list of aggregates for a value column
+            # Every value column has to have an entry
+            agg_func = agg_dict[pval] if len(agg_dict) > 1 else \
```
iyerr3: Use the `agg_dict` directly as a bool to check if it's empty.

orhankislal: agg_dict is not empty. It has either a single record (the default) or a list for each value column.

iyerr3: I see. I got confused with this. So a user can't add some values in the agg_dict expecting default for the rest?

iyerr3: also why do we add `__madlib_pivot_def_agg__` to `agg_dict` and then check for `len > 1` instead of just checking for empty dict?

orhankislal: No, they either use a function(s) for every value column or provide a value to function(s) mapping.

iyerr3: To understand the behavior: if a user does not specify the `aggregate_func` parameter then we use a default. If this is a list then we apply that for all values. But it's not possible to say: use default for `val1` and use `val2=[a, b, c]`?

orhankislal: I added the default to the dictionary to keep track of one less variable name and keeping all of aggregate functions in the same list seemed like an acceptable idea. 

Yes that is the behavior. I can change the behavior to assume the default function (avg) for the missing value columns but that will bring a new issue where the user tries to define the default as well as a list.

iyerr3: Also I now see that `__madlib_pivot_def_agg__` isn't just the default agg but can also be the list of input agg. That is going to confuse folks in the future. 

How about this: we have a single variable `agg_func_map` that is always a dict and contains mapping between value and aggregate list to apply. Depending on input (`aggregate_func`) we populate this mapping as follow: 
1. if input is NULL then each `val` is mapped to `[avg]`
2. if input is an agg or a list of aggs then each `val` is mapped to `list | [agg]`
3. if input is a mapping then we start with 1 and update mapping for each `val` in input. 

All of this logic can be in a separate function with good explanation. That way the main function remains simple: for each value we have a mapping that gives the agg to apply on that value.

orhankislal: I know it looks a bit overcomplicated, that is why I added the comments. You suggestion will look much cleaner. The only issue is we have to create a map that has every value column to the same aggregate(s) if the user doesn't need a map. It shouldn't be that costly since 1600 is still the upper limit.

iyerr3:
Could you please change all dynamically created SQL strings to have the query structure as a single string with variables filling in the dynamic entries? It's much easier to understand the query as a single string. Let me know if you need specific pointers.

orhankislal:
Out of 5 sql queries, I reduced the first 3 to single lines as you suggested. For the other two I would rather keep them broken into a list since they are created inside the 4 level loop.

iyerr3:
Even for the other two, I suggest keeping the static parts all together with only the dynamic part created in the loops. That avoids having the `SELECT` clause separated from the `FROM` by 70 lines. That avoids creating a variable `insert_sql_begin` 50+ lines from where it's used (and can be replaced by static string `INSERT INTO`).

orhankislal:
I see what you mean, should be easy enough.

PR #56, opened 2016-07-14
Path: Add support for overlapping patterns

mktal:


src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -155,15 +158,26 @@ def path(schema_madlib, source_table, output_table, partition_expr,
...
+            else:
+                plpy.execute("""
+                    CREATE TABLE {output_table} AS
+                       SELECT
+                            {p_col_orig_name_str},
+                            {match_id_name},
+                            {agg_func}
+                       FROM {matched_rows}
+                       GROUP BY {partition_expr}, {match_id_name}
+                    """.format(**locals()))
```
njayaram2: I presume this is not something specific for overlapping patterns, right?

I guess we can now remove "as_partition_expr" from this function, and
the else clause a little earlier in the function:
if not partition_expr:
            partition_expr = "1 = 1"
            as_partition_expr = "1 = 1 as all"
else:
            as_partition_expr = partition_expr

src/ports/postgres/modules/utilities/path.sql_in
```diff
@@ -531,7 +548,7 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.path(
     symbol_expr         VARCHAR,
     pattern_expr        VARCHAR
 ) RETURNS TEXT AS $$
-    SELECT MADLIB_SCHEMA.path($1, $2, $3, $4, $5, $6, NULL, NULL)
+    SELECT MADLIB_SCHEMA.path($1, $2, $3, $4, $5, $6, NULL, NULL, NULL)
 $$ LANGUAGE SQL
 m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `MODIFIES SQL DATA', `');
```
njayaram2: Having an example with overlapping_patterns = TRUE might be useful in the
user doc. The same example showing the output with and without overlapping
patterns might be helpful.

src/ports/postgres/modules/utilities/path.py_in
mktal: Good catch. I removed it in the latest commit that also included updates on docs.

src/ports/postgres/modules/utilities/path.py_in
```diff
@@ -40,7 +40,7 @@ m4_changequote(`<!', `!>')
 
 def path(schema_madlib, source_table, output_table, partition_expr,
          order_expr, pattern_expr, symbol_expr, agg_func,
-         persist_rows=None, **kwargs):
+         persist_rows=None, overlapping_patterns=None, **kwargs):
```
njayaram2: It might be a good idea to set the default value of overlapping_patterns to False instead of None. 
From what I gather, it is better to do that in the path.sql_in file where you can pass False instead of 
NULL in the overloaded function for the overlapping_patterns param. 

Of course, we will still need to have the "if overlapping_patterns is None:" check below to handle 
the case where overlapping_patterns is explicitly set to NULL by the user.

src/ports/postgres/modules/utilities/path.py_in
njayaram2: LGTM.

src/ports/postgres/modules/utilities/path.py_in
mktal: I don't see why setting the default value to False is better. As you said, the user can still explicitly set to NULL and we still need to account for that case.

njayaram2: I had received a similar comment on another module (sessionize), so thought it might be applicable here. @iyerr3 do you think setting the default to False is something we should do here too?

mktal:
@njayaram2 I removed default values in python and changed default values of boolean from `NULL` to `False` in sql file. Now in python we can just write `not partition_expr` to check if it is `True` or `False (None)`, since there is no need for now to differentiate between `False` and `None`

njayaram2:
LGTM.

PR #57, opened 2016-07-19
MADLIB-944: Add PostgreSQL 9.5 and 9.6 support

snaga:
* Add PostgreSQL 9.5 support.
* Fix to remove madlib.array_agg() in 9.5 or later.
* Add PostgreSQL 9.6 support.

https://issues.apache.org/jira/browse/MADLIB-944

### DIFF
--- a/methods/array_ops/src/pg_gp/array_ops.sql_in
+++ b/methods/array_ops/src/pg_gp/array_ops.sql_in
@@ -608,7 +608,7 @@ BEGIN
   -- array_agg() compatibility function is only needed PostgreSQL 9.4 or earlier
   -- because 9.5 or later has pg_catalog.array_agg(anyarray).
   --
-  IF regexp_replace(version(), '.*PostgreSQL (\d+\.\d+).*', '\1')::float8 < 9.5 THEN
+  IF current_setting('server_version_num')::int < 90500 THEN
     CREATE AGGREGATE MADLIB_SCHEMA.array_agg( anyelement) (
        SFUNC     = array_append,
        STYPE     = anyarray
--- a/methods/array_ops/src/pg_gp/array_ops.sql_in
+++ b/methods/array_ops/src/pg_gp/array_ops.sql_in
@@ -602,11 +602,22 @@ m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `NO SQL', `');
  *        in later GPDB and Postgres versions.
  */
 DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.array_agg(anyelement) CASCADE;
-CREATE AGGREGATE MADLIB_SCHEMA.array_agg( anyelement) (
-   SFUNC     = array_append,
-   STYPE     = anyarray
-   m4_ifdef( `__POSTGRESQL__', `', `, PREFUNC   = array_cat')
-);
+DO $$
+BEGIN
+  --
+  -- array_agg() compatibility function is only needed PostgreSQL 9.4 or earlier
+  -- because 9.5 or later has pg_catalog.array_agg(anyarray).
+  --
+  IF regexp_replace(version(), '.*PostgreSQL (\d+\.\d+).*', '\1')::float8 < 9.5 THEN
+    CREATE AGGREGATE MADLIB_SCHEMA.array_agg( anyelement) (
+       SFUNC     = array_append,
+       STYPE     = anyarray
+       m4_ifdef( `__POSTGRESQL__', `', `, PREFUNC   = array_cat')
+    );
+  END IF;
+END
+$$
+LANGUAGE 'plpgsql';
 
 /**
  * @brief Function to check if array contains NULL.
--- a/src/ports/postgres/9.5/CMakeLists.txt
+++ b/src/ports/postgres/9.5/CMakeLists.txt
@@ -1,0 +1,2 @@
+add_current_postgresql_version()
+add_extension_support()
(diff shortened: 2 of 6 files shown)


methods/array_ops/src/pg_gp/array_ops.sql_in
```diff
@@ -602,11 +602,22 @@ m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `NO SQL', `');
...
-   STYPE     = anyarray
-   m4_ifdef( `__POSTGRESQL__', `', `, PREFUNC   = array_cat')
-);
+DO $$
+BEGIN
+  --
+  -- array_agg() compatibility function is only needed PostgreSQL 9.4 or earlier
+  -- because 9.5 or later has pg_catalog.array_agg(anyarray).
+  --
+  IF regexp_replace(version(), '.*PostgreSQL (\d+\.\d+).*', '\1')::float8 < 9.5 THEN
```
decibel: It would be more straightforward to do current_setting('server_version_num')::int>=90500

snaga: @decibel Looks better. Will replace with it. Thanks!

iyerr3: I'm confused about this change and why we even need this function. Older versions of GPDB did not have array_agg (since they were based on PG 8.3) and needed this function. All currently supported platforms have this functionality. I realize 9.5 also supports array_agg(anyarray), but this function was not intended to provide that feature.

snaga: If we can drop this function at all, it would make things simpler. Does the latest GPDB have array_agg(anyarray) function? I'm not familiar with GPDB.

I'm not sure MADlib could work with 9.4 without this function, but at least, we need to drop this function for 9.5 or later because the function signature causes a conflict error with the built-in one.

BTW, is there any MADlib docs mentioning "current supported platform"? I could not find it...

danielgustafsson:
The current version of GPDB (4.3.8.x) does not have `array_agg(anyarray)` and it's not included in the upcoming 5.0 codebase either.

fmcquillan99:
Here are the supported platforms for MADlib 1.9:
https://cwiki.apache.org/confluence/display/MADLIB/FAQ#FAQ-Q1-2WhatplatformsdoesMADlibsupport?

Will update this for 1.9.1 to include Postgres 9.5, 9.6

Typically the last two versions of each platforms are supported with each MADlib release.

snaga:
Hi, I updated the PR and confirmed that "install-check" can pass on 9.4, 9.5 and 9.6beta2.
Please take a look.

vadv:
Hello!
After build and `make install` with this patch, I can't install madlib on PostgreSQL 9.5:
```
madpack.py : INFO : Detected PostgreSQL version 9.5.
madpack.py : INFO : *** Installing MADlib ***
madpack.py : INFO : MADlib tools version    = 1.9 (/usr/local/madlib/Versions/1.9/bin/../madpack/madpack.py)
madpack.py : INFO : MADlib database version = None (host=localhost:5432, db=gis, schema=madlib)
madpack.py : INFO : Testing PL/Python environment...
madpack.py : INFO : > Creating language PL/Python...
madpack.py : INFO : > PL/Python environment OK (version: 2.7.12)
madpack.py : INFO : Installing MADlib into MADLIB schema...
madpack.py : INFO : > Creating MADLIB schema
madpack.py : INFO : > Creating MADLIB.MigrationHistory table
madpack.py : INFO : > Writing version info in MigrationHistory table
madpack.py : INFO : > Creating objects for modules:
madpack.py : INFO : > - array_ops
madpack.py : ERROR : Failed executing /tmp/madlib.78jPYf/array_ops/array_ops.sql_in.tmp
madpack.py : ERROR : Check the log at /tmp/madlib.78jPYf/array_ops/array_ops.sql_in.log
madpack.py : INFO : Rolling back the installation...
madpack.py : INFO : Rollback finished successfully.
madpack.py : ERROR : MADlib installation failed.
```
Then I change m4 macros in file /usr/local/madlib/Versions/1.9/ports/postgres/9.5/madpack/SQLCommon.m4:
```
-__DBMS_ARCHITECTURE__=
+m4_define(`__DBMS_ARCHITECTURE__', `x86_64')
```
installation working for me.

snaga:
Hi @vadv, thanks for the feedback.
I guess something around cmake may have an issue in your case because `SQLcommon.m4` is generated by cmake during the build.
Could you please check and share your `build/src/ports/postgres/9.5/madpack/SQLCommon.m4` for more information?

snaga:
Anyway, I always do `make package` and `rpm -i` to install MADlib, so I haven't tried `make install` actually.

iyerr3:
LGTM. 

Both 9.4 and 9.5 worked without the madlib.array_agg function - I suggest we delete it as part of this commit. None of the MADlib functions use it anyways (except LDA which was been fixed here). 

@snaga I can do the deletion along with the merge if you're OK.

snaga:
@iyerr3 I think the code of madlib.array_agg() function is shared among PostgreSQL and GPDB. PostgreSQL (9.3~9.6) can work without this function, but GPDB still needs this function. Thus, it needs to be kept in this source for GPDB, doesn't it?

iyerr3:
@snaga MADlib is supported for GPDB 4.2 and 4.3 and both versions have the `array_agg(anyelement)` function. 

I'll delete the function and merge this PR.

PR #60, opened 2016-07-29
Path: Proposed fix for JIRA MADLIB-1012

mktal:


iyerr3:
👍  I guess we were looking at the same thing. I already pushed the fix in da582e8. You can close this one when you get the chance.

fmcquillan99:
I guess the wider question is are we supporting standard postgres identifiers for symbols? 
https://www.postgresql.org/docs/8.2/static/plpgsql-statements.html 

So far I have assumed yes

iyerr3:
Yes the expected behavior is that all standard postgresql identifiers are
allowed.

On Fri, Jul 29, 2016 at 3:15 PM, Frank McQuillan <notifications@github.com>
wrote:

> I guess the wider question is are we supporting standard postgres
> identifiers for symbols?
> https://www.postgresql.org/docs/8.2/static/plpgsql-statements.html
>
> So far I have assumed yes
>



-- 

---------------------------------------------------------
Rahul Iyer
Principal software engineer | Predictive Analytics

*Pivotal**A new platform for a new era*

mktal:
Great. I will go ahead and close this one.

PR #64, opened 2016-08-17
Release v1.9.1: Adds support for upgrading from 1.x to 1.9.1

orhankislal:


iyerr3:
The changes would have been easier to read if we had done `git mv changelist.yaml changelist_1.8_1.9.1.yaml` and then added a new `changelist.yaml` (or maybe you did that and git ignored those steps?).

orhankislal:
I don't think that would help since the moved file (`changelist_1.8_1.9.1.yaml`) is changed after the move.

iyerr3:
Yes, but those would only be the 1.9 -> 1.9.1 changes and not the 1.8 -> 1.9 changes that show as deleted in one file and added back in another file.

PR #65, opened 2016-08-22
Pivoting

orhankislal:
Adds filter for removing nulls from the aggregated results.

### DIFF
--- a/src/ports/postgres/modules/utilities/pivot.py_in
+++ b/src/ports/postgres/modules/utilities/pivot.py_in
@@ -204,8 +204,7 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
                     # Prepare the entry for the dictionary
                     insert_str.append("(\'__p_{dict_counter}__\', \'{pval}\', "
                                       "\'{agg}\' ".format(dict_counter=dict_counter,
-                                                          pval=pval,
-                                                          agg=agg))
+                                                          pval=pval, agg=agg))
 
                 # For every pivot column in a given combination
                 for counter, pcol in enumerate(pcols):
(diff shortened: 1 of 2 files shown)


src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -234,30 +235,46 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
...
-                pivot_str = ("{fill_str_begin}"
-                             "{agg} (CASE WHEN {pivot_begin} THEN {pval} END) "
-                             "{fill_str_end} AS {pivot_end}".
-                             format(fill_str_begin=fill_str_begin,
+
+                # Create the masked set
+                pivot_str_from = ("(CASE WHEN {pivot_begin} THEN {pval} END) "
+                            "AS {pivot_end}".
+                            format(fill_str_begin=fill_str_begin,
+                                    fill_str_end=fill_str_end,
```
iyerr3: We don't need the additional format specifiers

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -234,30 +235,46 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
...
+                            "{agg} ({pivot_end}) FILTER ("
+                            "WHERE {pivot_end} IS NOT NULL) "
+                            "{fill_str_end} AS {pivot_end}".
+                            format(fill_str_begin=fill_str_begin,
                                     fill_str_end=fill_str_end,
                                     pval=pval, agg=agg,
                                     pivot_begin=' AND '.join(pivot_str_begin),
                                     pivot_end=''.join(pivot_str_end)))
-                pivot_out_list.append("," + pivot_str)
+                pivot_str_sel_list.append("," + pivot_str_sel)
```
iyerr3: Would be nicer to only add `pivot_str_sel` in the list and let `join` take care of adding the commas.

orhankislal:
Thanks for the comments @iyerr3. I made the changes to clean up the code.

src/ports/postgres/modules/utilities/test/pivot.sql_in
```diff
@@ -152,15 +152,19 @@ SELECT assert(__p_7__ = 1.5,
...
 
-DROP AGGREGATE IF EXISTS array_accum (anyelement);
-CREATE AGGREGATE array_accum (anyelement)
-(
-    sfunc = array_append,
+DROP FUNCTION IF EXISTS array_add1(ANYARRAY, ANYELEMENT);
+DROP AGGREGATE IF EXISTS array_accum1 (anyelement);
+CREATE FUNCTION array_add1(ANYARRAY, ANYELEMENT) RETURNS ANYARRAY AS $$
+  SELECT $1 || $2
+$$ LANGUAGE sql STRICT;
```
decibel: It would be best to also mark this as IMMUTABLE so the executor can inline it. (It would be even better to use the built-in array_agg if it exists; it's probably even faster.)

decibel: Oh, just noticed this is just a test file, so guess it probably doesn't matter...

orhankislal: I still appreciate the pointer. This test exists to ensure user defined aggregates work with pivoting in addition to the built-in ones.

PR #69, opened 2016-09-30
Elastic Net: Add grouping support

njayaram2:
JIRA: MADLIB-950

Changes to add grouping support for Gaussian and Binomial models
that use FISTA for optimization. This has to be replicated to using
IGD for optimization, which will be done after getting feedback
on the changes made in this commit.

IMPORTANT: There is a parameter called "use_active_sets", and I am
not sure what the implications of that parameter are. Deferring
any changes to handle that for later.

Goal of this PR is to get feedback on Fista optimization.

@iyerr3 @mktal


src/ports/postgres/modules/elastic_net/elastic_net.py_in
```diff
@@ -369,6 +378,37 @@ def elastic_net_train(schema_madlib, tbl_source, tbl_result, col_dep_var,
...
+        cols_in_tbl_valid(source_table,
+                          _string_to_array_with_quotes(grouping_col),
+                          'elastic_net')
+        intersect = frozenset(
+            _string_to_array(grouping_col)).intersection(
+                frozenset(
+                    ('coef', 'random_feature_data',
+                     'random_feature_data', 'loss'
+                     'num_rows_processed', 'num_rows_skipped',
+                     'norm_of_gradient', 'num_iterations')))
```
njayaram2: Must discuss what must go in the frozenset for elastic net.

src/ports/postgres/modules/convex/utils_regularization.py_in
```diff
@@ -65,6 +65,12 @@ def __utils_dep_var_scale(**kwargs):
     return y_scale
 # ========================================================================
 
+def _cast_if_null(input, alias):
```
njayaram2: Moved this function to utilities.utilities.

src/ports/postgres/modules/utilities/in_mem_group_control.py_in
```diff
@@ -54,6 +54,18 @@ class BaseState(object):
...
 
+    def get_last_state_value_for_elasticnet(self):
+        ## Naming this function as something specific to elastic net
+        ## because this behavior might not be what a different 
+        ## algorithm might need out of obtaining the last state value.
+        ## The last element in value[] represents the log_likelihood
+        ## of a group-specific state, for Elastic Net. If the log_likelihood
+        ## of even a single group is not 0, we return 1. 
+
+        if any([value[len(value)-1] for value in self._state.itervalues()]):
```
njayaram2: Renamed the function as suggested.

njayaram2:
Thank you for all the comments @iyerr3 . I have incorporated your comments in the latest commit, and responded to them individually here as well.

src/ports/postgres/modules/elastic_net/elastic_net_generate_result.py_in
```diff
@@ -13,50 +15,131 @@ def __elastic_net_generate_result (optimizer, iteration_run, **args):
...
+                 standardize       boolean,
+                 iteration_run     integer)
+             """.format(select_grouping_info=select_grouping_info, **args))
+
+    result = plpy.execute(out_table_qstr)
+    for res in result:
+        build_output_table(res, grouping_column, grouping_str1,
+            standardize_flag, iteration_run, **args)
 
+    ## Create summary table, listing the grouping columns used.
```
iyerr3: Are there other statistics that can be added to the summary table (similarly to say GLM)?

src/ports/postgres/modules/utilities/in_mem_group_control.py_in
```diff
@@ -54,9 +54,19 @@ class BaseState(object):
                    ret_states=None, **kwargs):
         self.update(col_grp_key, col_grp_state, ret_states)
 
+    def get_last_state_value(self):
```
iyerr3: Can we refactor to make this a boolean output? It looks like it checks if the last element is "set" for each group. So something like `and(value[-1] for ...)` is more appropriate. 

If the group state values are float then having a boolean check might not be ideal. An `almostEqual to 0` check is better. Even with an `int` value, I prefer to see a `== 0` check to make the intention explicit.

Finally, I would suggest changing the name since, technically, it's not "getting" the last state value - I suggest a name along the lines of `are_last_state_value_zero`, with a `True` if all are `0` and `False` otherwise.

src/ports/postgres/modules/utilities/in_mem_group_control.py_in
```diff
@@ -54,9 +54,19 @@ class BaseState(object):
...
+        ## of every group is 0. If the value is non-zero even for a single
+        ## group, it returns a 1.
+        if any([value[-1] for value in self._state.itervalues()]):
+            return 1
+        return 0
+
     def update(self, col_grp_key, col_grp_state, ret_states):
         failed_grp_keys = []
-        if ret_states is None:
+        if ret_states is None or len(ret_states) == 0:
```
iyerr3: this is same as `if not ret_states`

src/ports/postgres/modules/elastic_net/elastic_net.py_in
```diff
@@ -34,220 +41,345 @@ def elastic_net_help(schema_madlib, family_or_optimizer=None, **kwargs):
...
+                      to the step size. Sometimes, this can speed
+                      up the calculation.
+
+    When warmup is True and warmup_lambdas is NULL, warmup_lambda_no
+    of lambda values will be automatically generated and used.
+
+    Reference --------------------------------
+    [1] Beck, A. and M. Teboulle (2009), A fast iterative
+        shrinkage-thresholding algorothm for linear inverse
+        problems. SIAM J. on Imaging Sciences 2(1),, 183-202.
```
njayaram2: That was probably a typo, will remove it.

src/ports/postgres/modules/elastic_net/elastic_net_generate_result.py_in
njayaram2: These were the statistics that were already present in elastic net. Adding more stats to the summary table might be possible, but would that warrant a broader discussion?

src/ports/postgres/modules/utilities/in_mem_group_control.py_in
njayaram2: Thank you, changed this.

PR #70, opened 2016-10-17
Build: Fix version parsing in madpack

iyerr3:
Primary change is line 359. Other changes are whitespace fixes. 

The primary bug in `re.findall('[0-9]', rev)` was that a version string like `1.3.10` resulted in output `[1, 3, 1, 0]` instead of `[1, 3, 10]`. This PR aims to fix that.


iyerr3:
@fmcquillan99 This fix is necessary to correctly install on any platform that reports a sub-version number `10` or higher.

fmcquillan99:
@iyerr3 Thanks for catching this

src/madpack/madpack.py
```diff
@@ -357,7 +356,7 @@ def _get_rev_num(rev):
         @param rev version text
     """
     try:
-        num = re.findall('[0-9]', rev)
```
mktal: In that case we could use `num = re.findall('[0-9]+', rev)`

mktal: Seems that rev is from `match_details.group(1)` which only contains numbers. Feel free to ignore the above comments.

iyerr3: The primary bug in `re.findall('[0-9]', rev)` was that a version string like `1.3.10` resulted in output `[1, 3, 1, 0]` instead of `[1, 3, 10]`. This PR solely aims to fix that. Ideally I would like to use something like `distutils.version.LooseVersion` to compare versions, but we also need the version parsed in a list and I'm not aware if `LooseVersion` provides that.

iyerr3:
Need final approval to close this PR.

iyerr3:
See recent update to enable versions of the form "1.9.1-rc". Per semantic versioning, all identifier labels should be a suffix starting with a hyphen. This is considered a valid version string. 

The caveat is that we don't yet have: `1.9.1 > 1.9.1-rc` which is one of the requirements of semantic versioning.

iyerr3:
Apologies for the rebased branch. New changes include: 

```
Changes:
- Fix madpack to assume revision string can be any valid Semantic
  versioning scheme.
- Change MADlib version to 1.10.0-dev
- Update MADLIB_VERSION_STRING to use underscore instead of the possible
  hyphen since some downstream programs (like gppkg) don't accept hyphen
  in the version string.
```

Please review and comment.

stephendotcarter:
Returning elements as str does not work when comparing different versions on line and is causing the install to fail on Greenplum 4.3.10:

```
if _get_rev_num(match_details.group(1)) >= _get_rev_num('4.3.5'):
    return '4.3ORCA'
```

This evaluates as:
```
>>> ['4','3','10'] > ['4','3','5']
False
```

They need to be int for the comparison to work:
```
>>> [4,3,10] > [4,3,5]
True
```

iyerr3:
Thanks for the comment, @stephendotcarter. 
I thought I had changed that to int last week but clearly I didn't push to remote. Please check the updated code - I've also added some unittests at the end of the file for basic sanity check (`RUN_TESTS` will be changed to `False` before merging).

stephendotcarter:
There is another line where we need to convert back to str for concatenation:
```
https://github.com/iyerr3/incubator-madlib/blob/0e5aa9f330c706a3d1e715879bbfd792953f1834/src/madpack/madpack.py#L518
temp_schema = schema + '_v' + ''.join(str(_get_rev_num(dbrev)))
```
In testing this gives me:
```
>>> ver = [4,3,10]
>>> 'test' + ''.join(str(ver))
'test[4, 3, 10]'
```
I think the line should be:
```
>>> 'test' + ''.join(map(str, ver))
'test4310'
```

And following should use `_is_rev_gte`:
```
https://github.com/iyerr3/incubator-madlib/blob/0e5aa9f330c706a3d1e715879bbfd792953f1834/src/madpack/madpack.py#L323
if _get_rev_num(match_details.group(1)) >= _get_rev_num('4.3.5'):
```

Other than that I have not seen any other locations that call `_get_rev_num`

iyerr3:
Thanks! Updated code.

PR #71, opened 2016-10-21
update pca docs

fmcquillan99:
doc updates for PCA 

esp clarify use of sparse matrices and add more examples


fmcquillan99:
closing because just see the PCP needs edits too.

will combine the 2 in later pull request

PR #74, opened 2016-11-08
Correct the version used in MADlib rpm installation

njayaram2:
Installing MADlib from an RPM creates a soft link for madpack.
The version number in the soft link contained '_dev' instead of
'-dev', which was introduced when MADlib version was changed to
1.10.0-dev. gppkg does not like an '-' in the RPM file name,
while the semantic versioning requires a '-' in the MADlib version.

@iyerr3


iyerr3:
Merged with commit 82e56a4

iyerr3:
@njayaram2: Could you please close this PR (forgot to close it with the commit).

PR #75, opened 2016-11-14
SVM: Implement c++ functions for training multi-class svm in mini-batch

mktal:
This PR implements multi-class support vector machine and a new training mechanism using buffered mini-batch stochastic gradient descent (hence "in-memory" computations across epochs). Preliminary experiments demonstrate potential on greatly improving training speed and accuracy in MADlib. For example, training on mnist datasets (10 classes) for **1s** on HAWQ 2.0 half-rack DCA and the model is able to achieve over **90% accuracy** on holdout data. The training and testing set in this case consist of 46900 and 23100 images, respectively.


src/modules/convex/algo/igd.hpp
```diff
@@ -86,6 +101,32 @@ IGD<State, ConstState, Task>::merge(state_type &state,
 
 template <class State, class ConstState, class Task>
 void
+IGD<State, ConstState, Task>::mergeInPlace(state_type &state,
```
haying: Hi, great to see you are using the convex framework to do this. I am trying to understand why you need to add the transitionInPlace and mergeInPlace. As I read the code, you can achieve what you are doing by using the old transition and merge function but to use algo.incrModel anywhere you are using task.model right now.

Of course, now you don't need to call final function to update the model. But with that, now you cannot calculate the loss across the whole epoch using a fixed model. It may or may not be an issue given the concrete use case.

mktal: Hi Aaron, it is great to receive your comment so quickly!:) Yes, you are correct that I can re-use the old transition and merge function if we keep another copy of the model as algo.incrModel. I decided to get rid of incrModel and update task.model in place because there is no need, algorithm-wise, to compute the loss and gradient across the whole epoch since we dont use line-search or other global mechanism here. In fact, I dont even keep the gradient around since I dont see the point of exposing the gradient to the end user. In this way, we save two thirds of the space required by the state and can potentially solve larger problems. Now, I understand there is code redundancy in transitionInPlace and mergeInPlace and I am open to discussion whether we should just factor them into a separate class.

haying: I see. It makes sense. Both loss and gradient are kept for the purpose to decide the convergence and step size tuning, to the upper layer of internal code or end user. I'm sure what you do for that, I guess you still need to pass the gradient to the inter-iteration layer? or it's not necessary anymore for this approach?

I think code redundancy is not a huge deal in this short piece of code. The man issue is making people wonder about the difference between the functions and which one to use in the future. I would suggest at least some amount of comment to explain the difference before we have time to refactor it cleanly.

Great work!

mktal:
The SQL code used to obtain the above results (note that we assume two tables mnist_train and mnist_test already exist in the database):

```sql
-- training on mnist_train and store the model in mnist_model
drop table if exists mnist_model;
create table mnist_model as
select
	madlib.structure_svm_step(
                      ind_var::double precision[][],
                      dep_var::double precision[],
                      NULL::double precision[],
                      784::integer, -- number of features
                      10::integer, -- number of classes
                      8::integer, -- epoch
                      64::integer, -- batch_size
                      0.01::double precision, -- step size
                      0.001::double precision, -- regularization
                      1::double precision) as state
from mnist_train;

-- testing on mnist_test
select
	1 - miss/total as accuracy
from (
	select count(*)::float8 as miss
	from mnist_test, mnist_model
	where madlib.structure_svm_predict(state, ind_var) <> dep_var
) q, (select count(*)::float8 as total from mnist_test) q2
```

haying:
I am not so sure about the technical details about convergence behavior of this algorithm. But it seems to me that it could be beneficial to expose the gradient so that users can do convergence tests and stepsize/batchsize tuning more easily. Of course, it would be even better if we can do more like providing a warning if we observe the model may not converge, etc.

mktal:
This is a good point Aaron. In terms of convergence behavior, it has both benefit of mini-batch that iterates fast and large batch size that reduce the variance of the empirical objective. To see this, note that we do multiple epoch within each buffer and given enough epoch we solve that buffer accurately, which can in turn be seen as applying multiple update towards the minimizer of objective formulated from that buffer. That is the reason why we only need one evoke of UDA and it gives us pretty good solution already. I encourage you to test yourself. I will also run more experiments in the future.

Regarding the gradient, I think in this case it is not a very good convergence indicator since hinge loss is not a smooth function, e.g., such that when the loss does not change much the gradient is still far from zero.  And one might have overfitting issue if one really force the gradient to be very small. So in this case loss value might be a better indicator of good solution (not necessarily the optimal solution but one with small generalization error).

haying:
Hi Xiaocheng,
Thanks for the explanation. It is good to know that it usually converges without much tuning. I would definitely try it out.

mktal:
- You definitely need to be careful with the `step size`, or learning rate, to make sure that it is either small enough or decayed properly, like other variants of sgd. For example, we can decay step size for every call of  `structure_svm_step` but within `structure_svm_step` we keep it constant. `
- `epoch` should not be chosen too large otherwise we will overfit one buffer too much (< 10)
- `batch_size` does not affect the convergence too much and often 32, 64 or 128 are some of the common choices

haying:
I suggest to document these in user doc.

asfgit:
https://builds.apache.org/job/madlib-pr-build/186/

asfgit:
https://builds.apache.org/job/madlib-pr-build/204/

fmcquillan99:
Update on this PR since it has been open for awhile.  This was some good work by mktal to build a multi-class svm module.

The issue with the PR is that the mini-batching is embedded in the svm code, whereas we would prefer to add mini-batching as a general capability in the stochastic gradient descent framework, so that it can be used by other modules besides.

There is work in progress on this currently:

Add mini-batching to IGD framework
https://issues.apache.org/jira/browse/MADLIB-1048
for the next release.

When that JIRA is done, we can then finish:

Multi-class SVM with mini-batching
https://issues.apache.org/jira/browse/MADLIB-1037

which is related to this PR.  @rahiyer may have more color to add as well.

asfgit:
https://builds.apache.org/job/madlib-pr-build/335/

njayaram2:
@mktal can you please close this PR? This code is now merged with https://github.com/apache/madlib/pull/229.

njayaram2:
The mini-batch related code in this PR was merged with https://github.com/apache/madlib/pull/229. Multi-class SVM related code is yet to be merged.

kaknikhil:
Closing this PR because of inactivity. We will take care of this in a future PR

PR #78, opened 2016-12-05
Graph: SSSP

orhankislal:
JIRA: MADLIB-992

- Introduces a new module: Graph.
- Implements the single source shortest path algorithm (Bellman-Ford).

### DIFF
--- a/src/ports/postgres/modules/graph/sssp.py_in
+++ b/src/ports/postgres/modules/graph/sssp.py_in
@@ -43,7 +43,8 @@ m4_changequote(`<!', `!>')
 def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
 		edge_args, source_vertex, out_table, **kwargs):
 	"""
-    Single source shortest path function for graphs
+    Single source shortest path function for graphs using the Bellman-Ford
+    algorhtm [1].
     Args:
         @param vertex_table     Name of the table that contains the vertex data.
         @param vertex_id        Name of the column containing the vertex ids.
@@ -52,14 +53,19 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
         						named arguments of the form "name=value".
         @param source_vertex    The source vertex id for the algorithm to start.
         @param out_table   	    Name of the table to store the result of SSSP.
+
+    [1] https://en.wikipedia.org/wiki/Bellman-Ford_algorithm
     """
 
 	with MinWarning("warning"):
 
 		INT_MAX = 2147483647
+		EPSILON = 1.0E-06
 
 		message = unique_string(desp='message')
-		toupdate = unique_string(desp='toupdate')
+
+		oldupdate = unique_string(desp='oldupdate')
+		newupdate = unique_string(desp='newupdate')
 
 		params_types = {'src': str, 'dest': str, 'weight': str}
 		default_args = {'src': 'src', 'dest': 'dest', 'weight': 'weight'}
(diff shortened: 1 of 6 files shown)


orhankislal:
Hi @decibel, I was wondering if you would be interested in taking a look at this PR. I would appreciate your comments/suggestions.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,347 @@
...
+
+		plpy.execute(
+			""" CREATE TABLE {out_table} AS
+				SELECT {vertex_id}::INT AS {vertex_id},
+					CAST({INT_MAX} AS INT) AS {weight},
+					CAST({INT_MAX} AS INT) AS parent
+				FROM {vertex_table} {distribution} """.format(**locals()))
+
+		plpy.execute(
+			""" CREATE TEMP TABLE {message}(
```
decibel: Ok, after reading [2], I see why this is called messages.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,347 @@
...
+    	@param source_table	Name of the table that contains the SSSP output.
+        @param out_table	The vertex that will be the destination of the
+            				desired path.
+	"""
+	validate_get_path(sssp_table, dest_vertex)
+	cur = dest_vertex
+	cols = get_cols(sssp_table)
+	id = cols[0]
+	ret = [dest_vertex]
+	sql = "SELECT parent FROM {sssp_table} WHERE {id} = {cur} LIMIT 1"
```
decibel: This will almost certainly be faster if done in a recursive CTE. If for some reason that won't work, at least plan the query and re-use the plan.

decibel:
BTW, if weight was real, float or numeric you could actually use Infinity as a value.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,347 @@
...
+
+		distribution = m4_ifdef(<!__POSTGRESQL__!>, <!''!>,
+			<!"DISTRIBUTED BY ({0})".format(vertex_id)!>)
+		local_distribution = m4_ifdef(<!__POSTGRESQL__!>, <!''!>,
+			<!"DISTRIBUTED BY (id)"!>)
+
+		validate_graph_coding(vertex_table, vertex_id, edge_table,
+			edge_params, source_vertex, out_table)
+
+		plpy.execute(" DROP TABLE IF EXISTS {0},{1}".format(message,toupdate))
```
orhankislal: `message` and `toupdate` are not user inputs. They are created locally at line 63-64.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,347 @@
...
+
+		plpy.execute(
+			""" CREATE TABLE {out_table} AS
```
orhankislal: Would this be an improvement for the ease-of-use only or the performance as well? Even if we give that as an option, the current implementation of the algorithm requires a place to store the intermediate results. Since this is in the order of O(|V|) we would still need a table for it.

orhankislal:
@decibel Thanks for your comments. Let me try to address some of them.
Number of updates are bound by the number of vertices in the worst case. Since one of our supported platforms (greenplum) has a 1 GB limit for any given field I am somewhat hesitant in this method. I will give it a try though. 
The `distributed by` clauses guide the database system for the indices. I agree that they would be useful for postgres.
I will try the edge representation with pgplsql to see how it affects the performance.

src/ports/postgres/modules/graph/sssp.py_in
decibel: Well, the issue exists throughout the code. Maybe it's just not worth worrying about... this code can only be called from inside the database anyway, right? If you're already in the database, you don't need SQL injection to break things, UNLESS the function is SECURITY DEFINER (which presumably this is not).

decibel:
Yeah, the 1GB limit is certainly a consideration. Using int4's for everything, a composite of (srt, dest, weight) would be 13-16 bytes (3 * int4 = 12 + varlena and maybe alignment). With the worst case of 16 bytes, that would be 65k edges, so that's the maximum that could be updated at once.

Before coding a full implementation, could you produce pseudocode of the ideal minimum set-based implementation would be? In particular, I'm thinking that {messages} is completely redundant, and if that's the case I suspect this entire algorithm could be done in a recursive CTE.

orhankislal:
I'll put the pseudocode in the JIRA (https://issues.apache.org/jira/browse/MADLIB-992)

orhankislal:
Unfortunately, GPDB does not support RECURSIVE option in WITH clause.

orhankislal:
Here is what I am thinking: We should create a new JIRA for SSSP Phase 2 to explore plpgsql implementation with an array to replace `toupdate` and `message` tables. I am not sure how this array will be distributed/broadcasted to the segments and would prefer to take a closer look at the performance measurements. 

In the meanwhile, we can combine these two tables with a simple workaround. The `message` table was necessary since we have truncate `toupdate` before repopulating it. If we use 2 `toupdate` tables, we can alternate them (fill one using the values of the other) and eliminate the need for `message`.

decibel:
Sounds good. I'd still like to see pseudocode as part of the 
documentation, because I don't think all these steps are necessary.

orhankislal:
@decibel I posted a pseudocode in the JIRA. Please let me know if that clarifies why I needed the multilayered sql code.

decibel:
You should really add that as a comment to the code. It's much clearer than reading the code itself.

And yes, it definitely helped me. I now realize what's been bugging me about this... the original algorithm on wikipedia is extremely general purpose; so much so that I missed an important fact: **a vertex should never need to be updated twice**, so long as you consider **all** edges of the vertex simultaneously. Of course, doing that is regular code would be a pain, but in SQL it's trivial.

Now that I understand what's going on, I see that the real trick here is to keep track of what the last set of updates was. I think it'd probably be a lot simpler to just keep a field that is what nesting level we're on, which is how this would end up coded in a CTE. I've tried a couple times to code that, but still haven't wrapped my head around it because of how far away from set theory the pseudo code is.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+			edge_params, source_vertex, out_table)
+
+		plpy.execute(" DROP TABLE IF EXISTS {0},{1},{2}".format(
+			message,oldupdate,newupdate))
+
+		plpy.execute(
+			""" CREATE TABLE {out_table} AS
+				SELECT {vertex_id}::INT AS {vertex_id},
+					CAST('Infinity' AS DOUBLE PRECISION) AS {weight},
+					CAST({INT_MAX} AS INT) AS parent
```
decibel: I think it'd be better to make that NULL instead of int_max. You can do that via `NULL::int AS parent`.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+						WHERE edge_table.{src}=oldupdate.id
+						GROUP BY edge_table.{dest}) x
+					WHERE edge_table.{src} = oldupdate.id
+						AND edge_table.{dest} = x.id
+						AND ABS(oldupdate.val + edge_table.{weight} - x.val) <
+							{EPSILON}
+					) AS message
+				WHERE message.id = out_table.{vertex_id}
+					AND message.val<out_table.{weight}
+					""".format(**locals()))
```
decibel: I don't think this WHERE and the DISTINCT are the right way to do this, but since I haven't been able to wrap my head around the intent here I don't know.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+				FROM
+				{oldupdate} AS oldupdate
+				WHERE
+				{out_table}.{vertex_id}=oldupdate.id
+				"""!>)
+			plpy.execute(sql.format(**locals()))
+
+			plpy.execute("TRUNCATE TABLE {0}".format(newupdate))
+			sql = ("""
+				INSERT INTO {newupdate}
```
decibel: This query desperately needs a comment block explaining in plain english what it's trying to accomplish. I'm pretty sure there's gains to be had here, but I'm not sure what it's trying to accomplish, so...

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+				SELECT DISTINCT ON (message.id) message.id AS id,
+					message.val AS val,
+					message.parent AS parent
+				FROM  {out_table} AS out_table,
```
decibel: Please use JOIN syntax. It's much clearer than separate WHERE clauses.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+					(SELECT edge_table.{dest} AS id, x.val AS val,
```
decibel: This subselect needs a comment block explaining what it's supposed to be doing.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -0,0 +1,372 @@
...
+						oldupdate.id AS parent
+					FROM {oldupdate} AS oldupdate, {edge_table} AS edge_table,
+						(SELECT edge_table.{dest} AS id,
```
decibel: This subselect needs a comment block explaining what it's supposed to be doing.

decibel:
Ok, I think I finally figured out what's going on here... you're keeping a table of every possible destination, as well as the minimum cost to that destination seen so far. The rest of this is just a question of walking through the edges, throwing away any paths that would exceed the cost to a destination that's already been seen, right?

orhankislal:
Yes that is the idea. I don't bother checking an edge unless its source vertex has been updated in the last pass.

src/ports/postgres/modules/graph/sssp.py_in
orhankislal: We need the DISTINCT ON since there might be multiple paths with the same cost. The inner-most query returns a single value (using the min function) but the middle query might match it to multiple edges. The algorithm doesn't guarantee a tie-breaking clause so whatever row DISTINCT ON chooses should be acceptable.

decibel:
On 12/19/16 1:55 PM, Orhan Kislal wrote:
> We need the DISTINCT ON since there might be multiple paths with the
> same cost. The inner-most query returns a single value (using the min
> function) but the middle query might match it to multiple edges. The
> algorithm doesn't guarantee a tie-breaking clause so whatever row
> DISTINCT ON chooses should be acceptable.

Please add comments to that effect to the code.

PR #80, opened 2016-12-13
KNN Added

auonhaidar:
I have added implementation of K-Nearest Neighbours. The function takes following arguments in order:
_training table name
training table's column name having vectors
training table's column name having target values(labels): assumed float
testing table name
testing table's column name having vectors
testing table's column name having row ids (used for final result's table creation)
Operation ('c' for classification, 'r' for regression)
K_

**sample run:** 
select * from madlib.knn('training_knn','points','label','testing_knn','points','pid','c',3);

**Output gets saved in knn_final table as:**

_pid, points, predicted_target_value_

Nearest neighbours are calculated using madlib's _squared norm_ metric.

### DIFF
--- a/deploy/gppkg/madlib.spec.in
+++ b/deploy/gppkg/madlib.spec.in
@@ -2,13 +2,13 @@
 %define __os_install_post %{nil}
 %define _rpmfilename      @MADLIB_GPPKG_RPM_FILE_NAME@
 %define _unpackaged_files_terminate_build 0
-%define _madlib_version  @MADLIB_VERSION_STRING@
+%define _madlib_version  @MADLIB_VERSION_STRING_NO_HYPHEN@
 
 BuildRoot:      @MADLIB_GPPKG_RPM_SOURCE_DIR@
 Summary:        MADlib for @GPDB_VARIANT@ Database
 License:        @CPACK_RPM_PACKAGE_LICENSE@
 Name:           madlib
-Version:        @MADLIB_VERSION_STRING@
+Version:        @MADLIB_VERSION_STRING_NO_HYPHEN@
 Release:        @MADLIB_GPPKG_RELEASE_NUMBER@
 Group:          @CPACK_RPM_PACKAGE_GROUP@
 Prefix:         /usr/local


src/ports/postgres/modules/knn/knn.sql_in
```diff
@@ -0,0 +1,126 @@
...
+        SELECT avg( $sql$ || label_column_name || $sql$ ) FROM  pg_temp.knn_interm $sql$
+        INTO l;
+        END IF;
+	EXECUTE 'INSERT INTO pg_temp.knn_label values($1,$2)'
+	USING cur_pid, l;
+    END LOOP;
+    
+    EXECUTE
+        $sql$
+	DROP TABLE IF EXISTS public.knn_final;
```
njayaram2: It might be better not to drop an existing output table. Instead,
we can either take the name of the output table as a parameter,
or we can append the string `"_output"` to the `test_source`
parameter. Of course, we still shouldn't drop the table, if the name
we end up constructing with that append results in a table name
that already exists. The function should throw an error saying the
output table already exists. This behavior will be in line with most
other MADlib functions.

src/ports/postgres/modules/knn/knn.sql_in
```diff
@@ -0,0 +1,126 @@
...
+    IF (k <= 0) THEN
+        RAISE EXCEPTION 'KNN error: Number of neighbors k must be a positive integer.';
+    END IF;
+    IF (operation != 'c' AND operation != 'r') THEN
+        RAISE EXCEPTION 'KNN error: put r for regression OR c for classification.';
+    END IF;
+    PERFORM MADLIB_SCHEMA.create_schema_pg_temp();
+    EXECUTE
+        $sql$
+    	DROP TABLE IF EXISTS pg_temp.knn_label;
```
njayaram2: I think this can be a temp table you create within the
madlib schema. Drop it before exiting.

src/ports/postgres/modules/knn/test/knn.sql_in
```diff
@@ -0,0 +1,6 @@
+m4_include(`SQLCommon.m4')
+/* -----------------------------------------------------------------------------
+ * Test knn.
+ *
+ * FIXME: Verify results
+ * -------------------------------------------------------------------------- */
```
njayaram2: Please go ahead and add few test cases, that would
check if knn is actually running as expected. Use small toy
data sets to do the same. We can run install checks to
perform some basic sanity checks of the module.

src/ports/postgres/modules/knn/knn.sql_in
```diff
@@ -0,0 +1,126 @@
...
+        INTO l;
+        END IF;
+	EXECUTE 'INSERT INTO pg_temp.knn_label values($1,$2)'
+	USING cur_pid, l;
+    END LOOP;
+    
+    EXECUTE
+        $sql$
+	DROP TABLE IF EXISTS public.knn_final;
+        CREATE TABLE public.knn_final AS
```
njayaram2: Why do we need both `pg_temp.knn_label` and `public.knn_final`?
Can't we just use the latter directly instead of knn_label?

src/ports/postgres/modules/knn/knn.sql_in
```diff
@@ -0,0 +1,126 @@
...
+    point_source VARCHAR,
+    point_column_name VARCHAR,
+    label_column_name VARCHAR,
+    test_source VARCHAR,
+    test_column_name VARCHAR,
+    id_column_name VARCHAR,
+    operation VARCHAR,
+    k INTEGER
+) RETURNS VARCHAR AS $$
+DECLARE
```
njayaram2: We can also overload this function. At the moment it looks
like only `k` can be an optional parameter. So we should be
able to run knn with some default value of `k`, if not provided.

Showing help messages when we call `knn()` or `knn('help')`
are examples of overloading this function.

njayaram2:
This is a great start! 
I will provide some github-specific feedback here, and more knn-specific
comments in the code.
Git can be daunting to use at first, but it's great once you get a hang of it.
I would recommend you go through the following wonderful book if you
have not already done so:
https://git-scm.com/book/en/v2

When you work on a feature/bug, it is best if you create a branch locally
and make all changes for that feature there. You can then push that branch
into your github repo and open a pull request. This way you won't mess with
your local master branch, which should ideally be in sync with the origin's
(apache/incubator-madlib in this case) master branch. More information on
how to work with branches can be found in the following chapter:
https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell 
(especially section 3.5)

One other minor feedback is to try including the corresponding JIRA id 
with the commit message. The JIRA associated with this feature is:
https://issues.apache.org/jira/browse/MADLIB-927

njayaram2:
Overall, this looks great and the code seems to be working correctly.
We can certainly make various modifications to it. For instance:
- Try to use other distance measures. If we can indeed use other distance
measures, we should have that as an optional parameter.
- We can try to improve the performance using parallel processing,
say on a distributed database like Greenplum. We can use
UDAs (with C++ code: Chapter 1 in http://madlib.incubator.apache.org/design.pdf
I had mentioned earlier) that can help in finding the k-nearest neighbors in parallel.
I guess that is something we can look at once you work on the comments
made on this version. We can discuss that when we get there!
- Incorporate other changes the community might suggest. There are
several postgres experts in the community who might be able to provide
suggestions to make the existing code more performant!

PR #82, opened 2016-12-21
Encode categorical variables

iyerr3:
JIRA: MADLIB-1038

Major overhaul of the dummy/one-hot encoding of categorical variables
with new name and updated arguments. Older function has been
deprecated with a warning to use the new function.


ivannovick:
nice

PR #87, opened 2017-01-13
Association Rules: Add rule counts to assoc_rules results

njayaram2:
JIRA: MADLIB-1044

The output column currently lists support, along with other columns.
This commit now lists the count too, which is essentially
support*number_of_transactions for any rule. This can certainly be
post-processed too, but listing it explicitly might be useful.

@orhankislal

### DIFF
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
@@ -589,7 +589,6 @@ INSERT INTO test_data VALUES (6, 'diapers');
 INSERT INTO test_data VALUES (6, 'chips');
 INSERT INTO test_data VALUES (7, 'beer');
 INSERT INTO test_data VALUES (7, 'diapers');
-\.
 
 Find all association rules with a support and threshold value of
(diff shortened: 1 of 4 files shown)


src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -504,3 +510,120 @@ def assoc_rules(madlib_schema, support, confidence, tid_col,
...
+INSERT INTO test_data VALUES (3, 'diapers');
+INSERT INTO test_data VALUES (4, 'beer');
+INSERT INTO test_data VALUES (4, 'chips');
+INSERT INTO test_data VALUES (5, 'beer');
+INSERT INTO test_data VALUES (6, 'beer');
+INSERT INTO test_data VALUES (6, 'diapers');
+INSERT INTO test_data VALUES (6, 'chips');
+INSERT INTO test_data VALUES (7, 'beer');
+INSERT INTO test_data VALUES (7, 'diapers');
+\.
```
orhankislal: I don't thin this is needed with insert into commands.

orhankislal:
LGTM +1

PR #89, opened 2017-01-23
K-means: support for array input

orhankislal:
JIRA: MADLIB-1018

Adds support for array input as data points. The function collates
the columns into a column in a temp table.

### DIFF
--- a/src/ports/postgres/modules/kmeans/kmeans.py_in
+++ b/src/ports/postgres/modules/kmeans/kmeans.py_in
@@ -19,6 +19,7 @@ from utilities.control_composite import IterationControllerComposite
 from utilities.validate_args import table_exists
 from utilities.validate_args import columns_exist_in_table
 from utilities.validate_args import table_is_empty
+from utilities.validate_args import is_var_valid
 from utilities.utilities import unique_string
 
 STATE_IN_MEM = m4_ifdef(<!__HAWQ__!>, <!True!>, <!False!>)
@@ -38,22 +39,30 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
 # ----------------------------------------------------------------------
 
 def kmeans_validate_expr(schema_madlib, rel_source, expr_point, **kwargs):
-    if not columns_exist_in_table(rel_source, [expr_point]):
+    """
+    Validation function for the expr_point parameter
+    expr_point accepts 2 formats:
+        - A single column name
+        - An array of column names
+    If the array notation is used, we don't check if the columns actually exist
+    because DB gives an appropriate message if the array notation isn't correct
+    """
 
-        p = re.compile('[Aa][Rr][Rr][Aa][Yy]\s*\[\s*[\s*\w|,\s*]+\s*\]')
+    if columns_exist_in_table(rel_source, [expr_point]):
+        return rel_source, False
 
-        if p.match(expr_point.strip()):
-            view_name = unique_string('km_view')
+    if is_var_valid(rel_source, expr_point):
+        view_name = unique_string('km_view')
 
-            plpy.execute(""" CREATE VIEW {view_name} AS
-                SELECT {expr_point} AS expr FROM {rel_source}
-                """.format(**locals()))
-            return view_name,True
-        else:
-            plpy.error(
-                """kmeans error: {expr_point} does not exist in {rel_source}!
-                """.format(**locals()))
-    return rel_source, False
+        plpy.execute(""" CREATE TEMP VIEW {view_name} AS
+            SELECT {expr_point} AS expr FROM {rel_source}
+            """.format(**locals()))
+        return view_name,True
+
+    else:
+        plpy.error(
+            """Kmeans error: {expr_point} is not a valid column or array!
+            """.format(**locals()))
 
 # ----------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
 
 # ----------------------------------------------------------------------
 
+def kmeans_validate_expr(schema_madlib, rel_source, expr_point, **kwargs):
```
decibel: This function could really use some comments or a docstring about what it's supposed to be doing.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
...
+    if not columns_exist_in_table(rel_source, [expr_point]):
+
+        p = re.compile('[Aa][Rr][Rr][Aa][Yy]\s*\[\s*[\s*\w|,\s*]+\s*\]')
```
decibel: A comment on what the regex is doing would be very helpful. The array part is pretty obvious (though, couldn't that just be done with an insensitive regex??), but the rest is more difficult to follow.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
 
 # ----------------------------------------------------------------------
 
+def kmeans_validate_expr(schema_madlib, rel_source, expr_point, **kwargs):
+    if not columns_exist_in_table(rel_source, [expr_point]):
```
decibel: I think it'd be much clearer to make this if columns_exist... and return immediately if true.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
...
+            plpy.execute(""" CREATE VIEW {view_name} AS
+                SELECT {expr_point} AS expr FROM {rel_source}
+                """.format(**locals()))
+            return view_name,True
+        else:
+            plpy.error(
+                """kmeans error: {expr_point} does not exist in {rel_source}!
+                """.format(**locals()))
+    return rel_source, False
+
```
decibel: It would be kinda nice to pre-compile the re by sticking a `kmeans_validate_expr.p = re.compile(...)` down here, as per http://stackoverflow.com/questions/279561/what-is-the-python-equivalent-of-static-variables-inside-a-function. Others might well have different ideas about that, though.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
...
 
+def kmeans_validate_expr(schema_madlib, rel_source, expr_point, **kwargs):
+    if not columns_exist_in_table(rel_source, [expr_point]):
+
+        p = re.compile('[Aa][Rr][Rr][Aa][Yy]\s*\[\s*[\s*\w|,\s*]+\s*\]')
+
+        if p.match(expr_point.strip()):
+            view_name = unique_string('km_view')
+
+            plpy.execute(""" CREATE VIEW {view_name} AS
```
decibel: Would this be better as a TEMP view? Or do all the callers create permanent tables?

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +37,25 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
...
+                SELECT {expr_point} AS expr FROM {rel_source}
+                """.format(**locals()))
+            return view_name,True
+        else:
+            plpy.error(
```
decibel: How does failure of the regex mean something doesn't exist in the database?

Maybe what you want is a `try:` block around the execute that returns this error, and a separate error if the regex doesn't match? If that's the case, I suggest changing the regex to `if not p.match(...): plpy.error()`.

orhankislal:
Hi @decibel, thanks for your comments. I removed the regex entirely and replaced it with a built-in madlib function. I addressed your other comments as well. Please let me know what you think.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -34,6 +38,34 @@ def kmeans_validate_src(schema_madlib, rel_source, **kwargs):
...
+        - An array expression
+    If the array expression is used, we use the is_var_valid function to
+    verify. This allows the regular array notation as well as any function
+    that returns an array.
+    """
+
+    if columns_exist_in_table(rel_source, [expr_point]):
+        return rel_source, False
+
+    if is_var_valid(rel_source, expr_point):
```
iyerr3: An alternative to this is to use `pg_typeof` to get the type of the expression. See `get_expr_type` in `validate_args.py_in` and check for '[]' in the result. 
This allows rejecting non-array expressions. 
@decibel any concerns with using `pg_typeof` here?

orhankislal: I think this is a better solution but we still allow non numeric arrays.

PR #91, opened 2017-01-24
Build: Use only major version for GPDB 5, HAWQ 2

iyerr3:
GPDB, starting 5.0, and HAWQ, starting 2.0, are using semantic
versioning for releases. This implies a binary compatibility between
same major versions. Hence, we only need to compile a single MADlib
binary for the same major version, with the folder name being just the
major version instead of 'major.minor'.


njayaram2:
LGTM +1

PR #93, opened 2017-01-25
Allow encode_categorical_variables() to use the svec type.

snaga:
Instead of the array.


fmcquillan99:
Thank you for the PR.

It seems you would like to always output an svec for this function?  i.e., not support regular array output?  This seems overly restrictive.

Most MADlib functions that would use the output from encode_categorical_variables (e.g., regression) accept array inputs but not necessarily svec inputs).

If we would like to support svec, I would suggest we would need to change the array_output param from BOOLEAN to multi-valued (no array, array, svec) to support the intent of this PR.

array_output (optional)
BOOLEAN. default: FALSE. This parameter controls the output format of the indicator variables. If FALSE, a column is created for each indicator variable. PostgreSQL limits the number of columns in a table. If the total number of indicator columns exceeds the limit, then make this parameter TRUE to combine the indicator columns into an array. The order of the array is the same as specified in 'categorical_cols'. A dictionary will be created when 'array_output' is TRUE to define an index into the array. The dictionary table will be given the name of the 'output_table' appended by '_dictionary'.

snaga:
Thanks for the review.

> If we would like to support svec, I would suggest we would need to change the array_output param from BOOLEAN to multi-valued (no array, array, svec) to support the intent of this PR.

Yes, it would be better to make array_output for multi-values.
I will try some modification to allow `column` (default), `array` and `svec` with changing the param name from `array_output` to `output_type`.

fmcquillan99:
Yes that sounds good.   So `output_type` will be text I assume.

iyerr3:
Thanks, Satoshi (@snaga). 
The changes look good - I merged this and updated the docs to reflect the new parameter.

PR #94, opened 2017-01-27
Use relative path for installation in GPDB/HAWQ

iyerr3:


iyerr3:
@orhankislal @njayaram2 Could you please verify this my installing on GPDB/HAWQ and checking if the path for C++, Python functions are using symlink. You can check the function location using `\df+ <func_name>`. Alternatively, you could use `-vl` during `madpack (re)install`, which will keep a `tmp` file for each module installation. Digging into any of the `tmp` files will let you know which path it's using. GPDB symlinks are called `greenplum-db` and HAWQ symlink is called `hawq`

orhankislal:
@iyerr3 I was able to confirm the symlink on gpdb and hawq with gppkg and rpm. LGTM +1

PR #95, opened 2017-02-02
Release v1.10:

orhankislal:
- Updates the changelists and other related files for upgrade.
Note that upgrade is not supported from versions prior to 1.8.

- Removes hardcoded schema name (madlib) from some of the tests.


rahiyer:
If the upgrade from versions prior to v1.8 is not supported then we should delete the corresponding changelist files to avoid confusion.

iyerr3:
If the upgrade from versions prior to v1.8 is not supported then we should delete the corresponding changelist files to avoid confusion.

orhankislal:
Sounds good, I'll add a warning as well. Actually, should we even allow the user to attempt an upgrade from old versions?

iyerr3:
IMO, if we're not supporting upgrade from prior to v1.8 then we should
force users to follow an upgrade chain.

PR #98, opened 2017-02-06
Bugfix/knn validation

njayaram2:
knn: Fix input validation issues

- Some input validation was missing which is included now.
- there was still a bug wrt ambiguous references to column names,
which was only partly handled previously.


orhankislal:
We should remove the hardcoded schema names from the tests as well. If MADlib is installed in a different schema, they might create problems.

orhankislal:
`knn.sql` lines 404 - 407 are not necessary anymore thanks to the expanded validation function.

PR #101, opened 2017-02-11
Multiple: Add casting to allow compilation with GCC 6+

iyerr3:
JIRA: MADLIB-1025

GCC 6+ introduced stricter rules for implicit casting where loss of
information is possible.

Closes #101


fmcquillan99:
Holding this for now for a bit more investigation, so won't merge just yet.

iyerr3:
Closing this for now. Will revisit this in future.

PR #103, opened 2017-02-15
Docker Build for MADlib

njayaram2:
JIRA: MADLIB-920

- Add docker files that would help developers download a docker image
with Postgres-9.6 and MADlib depedencies installed. A developer's
local source code changes can be built on this image's container
to quickly build and run install-checks. Requires docker installed
on the developer's environment.
- Add a bash script (jenkins_build.sh) that would be a starting
point towards getting a Jenkins build for MADlib master branch.
- This is work under heavy development and we would want to add in
similar support for Greenplum as well in the future. There is a
placeholder Dockerfile for GPDB in this commit, must be modified
to get it working.


iyerr3:
Looks good to me. I was able to run through the steps. Only comment would be to update documentation to reflect the `madlib` docker hub account.

PR #105, opened 2017-02-23
Graph:

orhankislal:
- Create generic graph validation and help message to standardize
future graph algorithm development.
- Expand the design document with more detail on the graph
representation as well as the SSSP implementation.

Closes #105


orhankislal:
Thanks for the comments @iyerr3. I tried to reorganize the algorithm explanation a bit, please let me know what you think.

src/ports/postgres/modules/graph/graph_utils.py_in
```diff
@@ -0,0 +1,102 @@
...
+----------------------------------------------------------------------------
+ SELECT {schema_madlib}.{func_name}(
+    vertex_table  TEXT, -- Name of the table that contains the vertex data.
+    vertex_id     TEXT, -- Name of the column containing the vertex ids.
+    edge_table    TEXT, -- Name of the table that contains the edge data.
+    edge_args     TEXT, -- A comma-delimited string containing multiple
+    			-- named arguments of the form "name=value".
+    {other_text}
+    out_table     TEXT  -- Name of the table to store the result of SSSP.
+);
```
njayaram2: Having a mandatory param such as `output_table` after `{other_text}` might not work well if we have optional params. It seems fine in sssp since other_text is essentially the starting vertex id (a mandatory param), but that might not be true for other modules. I suggest we move `other_text` to after `out_table`.

An example of a graph-based algorithm that uses optional params is PageRank. We will have optional params such as `max_iter` and `threshold` that will be listed after `out_table`.

orhankislal: I was trying to avoid changing the SSSP notation but I guess it is inevitable. Do you think separating `other_text` into two (`mandatory_params` and `optional_params`) could work for future graph algorithms?

njayaram2: Yes, that might work better. We can have `other_madatory_params` and `optional_params`, before and after `out_table` respectively. We may have to follow this rule for other graph modules: `out_table` must be our last mandatory param, to maintain some consistency. 

But this might not be in line with our existing modules. For example, I checked elastic_net and the output table is one of the mandatory params specified early on. There are several algorithm specific mandatory params following the output table name.
We should also put a comment in the code specifying the reason, else it will look confusing.

orhankislal: I think we might want to move `out_table` to the other parameters as well. For some functions like graph diameter, we don't have to create an output table. That will allow the pagerank to place its optional parameters after the `out_table`.

PR #106, opened 2017-02-24
Build: Fix module sort order for PGXN installation

iyerr3:
JIRA: MADLIB-1024

PGXN installation involves creating a single extension sql file that
contains all the SQL commands run during MADlib deployment. The modules
added into this extension file are to be placed in the right order,
taking dependencies into account.

MADlib has a function that compares a given file path with topologically
sorted modules to decide the order of concatenation to extension file.
This comparison is faulty since the module name was searched for in the
whole path, leading to false positive with modules that have another
module name as substring.  The specific bug was related to 'svec_util'
being flagged in same order as 'svec'.

This commit fixes this issue taking advantage of the file path names being
of the form '.../modules/<module_name>/...', hence comparing the
complete module name.

Closes #106

### DIFF
--- a/src/madpack/sort-module.py
+++ b/src/madpack/sort-module.py
@@ -1,7 +1,24 @@
 #!/usr/bin/env python
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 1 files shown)


orhankislal:
LGTM +1

PR #108, opened 2017-03-15
Pivot: Add support for array output

iyerr3:
JIRA: MADLIB-1066

When total pivoted columns exceed 1600, an array output becomes
essential. This commit adds support to get each pivoted set of columns
(all columns related to a particular value-aggregate combination) as an
array. There is also support for getting the output as madlib.svec.

### DIFF
--- a/src/ports/postgres/modules/utilities/pivot.py_in
+++ b/src/ports/postgres/modules/utilities/pivot.py_in
@@ -56,7 +56,7 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
         @param source_table     The original data table
         @param out_table        The output table that contains the dummy
                                 variable columns
-        @param index            The index columns to group by the records by
+        @param index            The index columns to group the records by
         @param pivot_cols       The columns to pivot the table
         @param pivot_values     The value columns to be summarized in the
                                 pivoted table
@@ -261,7 +261,7 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
 
                         # Collect pcol values for the dict
                         if output_col_dictionary:
-                            dict_insert_str.append("{0}".format(
+                            dict_insert_str.append("'{0}'".format(
                                 comb[counter] if comb[counter] is not None else "NULL"))
                     pivot_col_name.append('"')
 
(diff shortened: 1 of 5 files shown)


src/ports/postgres/modules/utilities/test/pivot.sql_in
```diff
@@ -149,7 +149,7 @@ SELECT pivot('pivset_ext', 'pivout', 'id, id2', 'piv, piv2', 'val, val2',
 SELECT * FROM pivout;
 
 SELECT assert(__p_7__ = 1.5,
```
orhankislal: Install-check fails at this point. Since the new dictionary starts from 1 (not 0) the __p_8__ column has this value.

src/ports/postgres/modules/utilities/test/pivot.sql_in
```diff
@@ -168,3 +168,17 @@ SELECT pivot('pivset_ext', 'pivout', 'id', 'piv', 'val', 'array_accum1');
 SELECT * FROM pivout;
 
 DROP VIEW IF EXISTS pivset_ext;
```
orhankislal: Drop view should happen at the end of the test file.

src/ports/postgres/modules/utilities/pivot.py_in
```diff
@@ -222,59 +261,78 @@ def pivot(schema_madlib, source_table, out_table, index, pivot_cols,
 
                         # Collect pcol values for the dict
                         if output_col_dictionary:
-                            insert_str.append("{0}".format(
+                            dict_insert_str.append("'{0}'".format(
```
orhankislal: Is there a particular reason this is needed? The install-check fails on NULL values and removing them seems to solve the issue.

src/ports/postgres/modules/utilities/pivot.sql_in
```diff
@@ -127,6 +134,18 @@ pivot(
     a dictionary output file will be created and a message given to the user.
     </dd>
 
+    <dt>output_type (optional)</dt>
+    <dd>VARCHAR. default: 'column'.  This parameter controls the output format
+    of the pivoted variables. If 'column', a column is created for each pivot
+    variable. PostgreSQL limits the number of columns in a table.
```
orhankislal: Maybe we should suggest a number for the user to start considering this option.

iyerr3:
ok to test

PR #109, opened 2017-03-22
Feature: PageRank

njayaram2:
JIRA: MADLIB-1069

- Introduces a new module that computes the PageRank of all nodes
in a directed graph.
- Implements the original PageRank algorithm that assumes a random
surfer model (https://en.wikipedia.org/wiki/PageRank#Damping_factor)

### DIFF
--- a/src/ports/postgres/modules/graph/pagerank.py_in
+++ b/src/ports/postgres/modules/graph/pagerank.py_in
@@ -47,14 +47,14 @@ def validate_pagerank_args(vertex_table, vertex_id, edge_table, edge_params,
     validate_graph_coding(vertex_table, vertex_id, edge_table, edge_params,
         out_table, module_name)
     _assert(damping_factor >= 0.0 and damping_factor <= 1.0,
-        """PageRank: Invalid damping factor value ({0}), must be between 0
-            and 1.""".format(damping_factor))
+        """PageRank: Invalid damping factor value ({0}), must be between 0 and 1."""
+        .format(damping_factor))
     _assert(threshold >= 0.0 and threshold <= 1.0,
-        """PageRank: Invalid threshold value ({0}), must be between 0
-            and 1.""".format(threshold))
+        """PageRank: Invalid threshold value ({0}), must be between 0 and 1."""
+        .format(threshold))
     _assert(max_iter > 0,
-        """PageRank: Invalid max_iter value ({0}), must be a positive integer.
-        """.format(max_iter))
+        """PageRank: Invalid max_iter value ({0}), must be a positive integer. """
+        .format(max_iter))
 
 def pagerank(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
     out_table, damping_factor, max_iter, threshold, **kwargs):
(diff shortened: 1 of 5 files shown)


src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -0,0 +1,290 @@
...
+m4_changequote(`<!', `!>')
+
+def validate_pagerank_args(vertex_table, vertex_id, edge_table, edge_params,
+        out_table, damping_factor, max_iter, threshold, module_name):
+    """
+    Function to validate input parameters for PageRank
+    """
+    validate_graph_coding(vertex_table, vertex_id, edge_table, edge_params,
+        out_table, module_name)
+    _assert(damping_factor >= 0.0 and damping_factor <= 1.0,
```
orhankislal: Should we add a similar boundary check for threshold as well?

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -0,0 +1,290 @@
...
+                WHERE {vertex_id} NOT IN (
+                    SELECT {vertex_id}
+                    FROM {message}
+                )
+            """.format(**locals()))
+        # Check for convergence
+        if 'e' in str(threshold):
+            thresh = abs(int(str(threshold).split('e')[-1]))
+        else:
+            thresh = len(str(threshold).split('.')[1])
```
orhankislal: 0.001 and 0.009 both give 3 but 0.009999 gives 6.

src/ports/postgres/modules/graph/pagerank.py_in
njayaram2: Will take care of this. Technically, any value of threshold would be legal so had skipped the test for that, but a value between 0 and 1 makes the most sense.

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -0,0 +1,290 @@
...
+        GROUP BY {src}
+        {cnts_distribution}
+        """.format(**locals()))
+
+    for i in range(max_iter):
+        plpy.execute("""
+                CREATE TABLE {message} AS
+                SELECT {edge_temp_table}.{dest} AS {vertex_id},
+                        SUM({v1}.pagerank/{out_cnts}.{out_cnts_cnt})*{damping_factor}+{random_prob} AS pagerank
+                FROM {cur}, {out_cnts}, {edge_temp_table}, {cur} AS {v1}
```
njayaram2: Will use joins instead of this, but not sure of the performance gain yet. Might have to revert if the results are not favorable.

src/ports/postgres/modules/graph/pagerank.py_in
njayaram2: Yes, and that's what I wanted it to be. The `ROUND(...)` function in the following query rounds off the page rank value to `thresh` number of decimal places. The difference between two page rank values (rounded off to the desired number of decimal points) should be less than less than `threshold` for convergence.

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -0,0 +1,290 @@
...
+                    FROM {message}
+                )
+            """.format(**locals()))
+        # Check for convergence
+        if 'e' in str(threshold):
+            thresh = abs(int(str(threshold).split('e')[-1]))
+        else:
+            thresh = len(str(threshold).split('.')[1])
+        if plpy.execute("""
+                SELECT (
```
njayaram2: Will take care of these convergence test comments as part of https://issues.apache.org/jira/browse/MADLIB-1082.

njayaram2:
Thanks for all the comments @orhankislal , will go ahead and merge this.

PR #112, opened 2017-04-06
Feautre: Add grouping support for PageRank

njayaram2:
MADLIB-1082

- Add grouping support for pagerank, which will compute a PageRank
probability distribution for the graph represented by each group.
- Add convergence test, so that PageRank computation terminates
if the pagerank value of no node changes beyond a threshold across
two consecutive iterations (or max_iters number of iterations are
done, whichever happens first). In case of grouping, the algorithm
terminates only after all groups have converged.
- Create a summary table apart from the output table that records
the number of iterations required for convergence. Iterations
required for convergence of each group is recorded when grouping
is used. This implementation also ensures that we don't compute
PageRank for groups that have already converged.

### DIFF
--- a/src/ports/postgres/modules/graph/pagerank.py_in
+++ b/src/ports/postgres/modules/graph/pagerank.py_in
@@ -31,33 +31,37 @@ import plpy
 from utilities.control import MinWarning
 from utilities.utilities import _assert
 from utilities.utilities import extract_keyvalue_params
-from utilities.utilities import unique_string
-from utilities.control import IterationController2S
+from utilities.utilities import unique_string, split_quoted_delimited_str
+from utilities.validate_args import columns_exist_in_table, get_cols_and_types
 from graph_utils import *
-
-import time
 
 m4_changequote(`<!', `!>')
 
-def validate_pagerank_args(vertex_table, vertex_id, edge_table, edge_params,
-        out_table, damping_factor, max_iter, threshold, module_name):
+def validate_pagerank_args(schema_madlib, vertex_table, vertex_id, edge_table,
+        edge_params, out_table, damping_factor, max_iter, threshold,
+        grouping_cols_list, module_name):
     """
     Function to validate input parameters for PageRank
     """
     validate_graph_coding(vertex_table, vertex_id, edge_table, edge_params,
         out_table, module_name)
     _assert(damping_factor >= 0.0 and damping_factor <= 1.0,
-        """PageRank: Invalid damping factor value ({0}), must be between 0 and 1."""
-        .format(damping_factor))
-    _assert(threshold >= 0.0 and threshold <= 1.0,
-        """PageRank: Invalid threshold value ({0}), must be between 0 and 1."""
-        .format(threshold))
+        """PageRank: Invalid damping factor value ({0}), must be between 0 and 1.""".
+        format(damping_factor))
+    _assert(not threshold or (threshold >= 0.0 and threshold <= 1.0),
+        """PageRank: Invalid threshold value ({0}), must be between 0 and 1.""".
+        format(threshold))
     _assert(max_iter > 0,
-        """PageRank: Invalid max_iter value ({0}), must be a positive integer. """
-        .format(max_iter))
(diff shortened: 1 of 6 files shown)


src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -158,44 +313,198 @@ def pagerank(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
+        that appear in cur_unvonv but not in message_unconv.
+        If this function is called after max_iter is completed, without
+        convergence, all the unconverged groups from cur_unconv is used
+        (note that message_unconv is renamed to cur_unconv before checking
+        for unconverged==0 in the pagerank function's for loop)
+    """
+    if message_unconv is None:
+        plpy.execute("""
+            DROP TABLE IF EXISTS {temp_summary_table};
+            CREATE TABLE {temp_summary_table} AS
```
orhankislal: We can overwrite the temp_summary_table variable instead of creating a new table (as long as we make sure not to drop it at the end).

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -158,44 +313,198 @@ def pagerank(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
-                    INNER JOIN {cur} AS {v1} ON {v1}.{vertex_id}={edge_temp_table}.{src}
-                GROUP BY {edge_temp_table}.{dest}
-            """.format(**locals()))
+                    INNER JOIN {cur} ON {cur_join_clause}
+                    INNER JOIN {out_cnts} ON {out_cnts_join_clause}
+                    INNER JOIN {cur} AS {v1} ON {v1_join_clause}
+                    {vertices_per_group_inner_join}
+                {ignore_group_clause}
+                GROUP BY {grouping_cols_select} {edge_temp_table}.{dest}
+            """.format(grouping_cols_select=edge_grouping_cols_select+', '
```
njayaram2: Sounds like a good idea, will change this.

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -158,44 +313,198 @@ def pagerank(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
         # If there are nodes that have no incoming edges, they are not captured in the message table.
         # Insert entries for such nodes, with random_prob.
         plpy.execute("""
                 INSERT INTO {message}
-                SELECT {vertex_id}, {random_prob}::DOUBLE PRECISION AS pagerank
-                FROM {cur}
-                WHERE {vertex_id} NOT IN (
+                SELECT {grouping_cols_select} {cur}.{vertex_id}, {random_jump_prob} AS pagerank
+                FROM {cur} {vpg_from_clause}
+                WHERE {vpg_where_clause} {vertex_id} NOT IN (
```
njayaram2: The query does takes care of the scenario you mentioned. The `{message_grp_where}` clause inside NOT IN checks if `cur.grouping_cols=message.grouping_cols`. I also tried an example to confirm it does work as expected. So leaving this as is.

src/ports/postgres/modules/graph/pagerank.py_in
```diff
@@ -158,44 +313,198 @@ def pagerank(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
+    plpy.execute("""DROP TABLE IF EXISTS {0},{1},{2},{3},{4},{5};
+        """.format(out_cnts, edge_temp_table, cur, message, cur_unconv,
+                    message_unconv))
+    if grouping_cols:
+        plpy.execute("""DROP TABLE IF EXISTS {0},{1},{2};
+            """.format(vertices_per_group, temp_summary_table,
+                distinct_grp_table))
     plpy.execute("SET client_min_messages TO %s" % old_msg_level)
 
+def update_result_tables(temp_summary_table, i, summary_table, out_table,
```
njayaram2: Using truncate instead of drop and create.

PR #113, opened 2017-04-07
Graph: Add grouping support to SSSP

orhankislal:
JIRA: MADLIB-1081

- This commit adds grouping support for SSSP as well as its path function.

### DIFF
--- a/src/ports/postgres/modules/graph/sssp.py_in
+++ b/src/ports/postgres/modules/graph/sssp.py_in
@@ -235,11 +235,11 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
 		if grouping_cols is not None:
 
 			distinct_grp_table = unique_string(desp='grp')
-			plpy.execute(""" DROP TABLE IF EXISTS {distinct_grp_table};
-                CREATE TEMP TABLE {distinct_grp_table} AS
-                SELECT DISTINCT ({grouping_cols}) FROM {edge_table}
-            """.format(**locals()))
-
+			plpy.execute(""" DROP TABLE IF EXISTS {distinct_grp_table} """.
+				format(**locals()))
+			plpy.execute( """ CREATE TEMP TABLE {distinct_grp_table} AS
+                SELECT DISTINCT {grouping_cols} FROM {edge_table} """.
+            	format(**locals()))
 			subq = unique_string(desp='subquery')
 
 			checkg_ds_sub = _check_groups(distinct_grp_table,subq,glist)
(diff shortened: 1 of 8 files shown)


src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -137,45 +231,131 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
...
+						{init_w} AS {weight},
+					NULL::INT AS parent
+					FROM (
+						SELECT {src} AS {vertex_id} {comma_grp}
+						FROM {edge_table} WHERE {gcomp}
+						UNION
+						SELECT {dest} AS {vertex_id} {comma_grp}
+						FROM {edge_table} WHERE {gcomp}
+						) x
+					WHERE {vertex_id} IS NOT NULL
```
njayaram2: I think it's better to use INNER JOINS instead of a `WHERE` clause.
This query does not work if the grouping columns are not integers.
For instance, if the grouping column is a `VARCHAR`, we will have to use
quoted values. It could get messy to construct the where clause based
on types of the grouping columns.

I created the following example for grouping, based off the user docs 
example in SSSP (not sure, but I presume this is a valid input table):
```
DROP TABLE IF EXISTS edge_gr,out_gr,out_gr_summary_out_gr_path;
CREATE TABLE edge_gr AS
(
  SELECT *, 'a' AS grp FROM edge
  UNION
  SELECT *, 'b' AS grp FROM edge WHERE src < 6 AND dest < 6
);
INSERT INTO edge_gr VALUES
(4,5,-20,'b');
```
Running the SSSP algorithm on this table results in an error.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -198,110 +378,228 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
...
+					GROUP BY {grouping_cols}""".format(**locals()))
+
+			for i in gvalues:
+				cur = dest_vertex
+				gcheck = []
+				gset = []
+				for j in glist:
+					gcheck.append("{0} = {1}".format(j,i[j]))
+					gset.append("{0}".format(i[j]))
+				gsql = " AND " + " AND ".join(gcheck)
```
njayaram2: Try using list comprehension here as well, it is supposed
to be faster than append.

src/ports/postgres/modules/graph/sssp.py_in
orhankislal: Thanks for catching this. I was able to fix it using the same method from your pagerank PR.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -137,45 +226,116 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
...
+		# The initialization step is quite different when grouping is involved
+		# since not every group (subgraph) will have the same set of vertices.
+
+		# Example:
+		# Assume there are two grouping columns g1 and g2
+		# g1 values are 0 and 1. g2 values are 5 and 6
+		if grouping_cols is not None:
+
+			distinct_grp_table = unique_string(desp='grp')
+			plpy.execute(""" DROP TABLE IF EXISTS {distinct_grp_table} """.
```
njayaram2: Some temp tables such as `distinct_grp_table`, `new update`, `old update` and `temp_table`  are not dropped at the end. Not sure if there are other such tables, please check and drop them.

PR #115, opened 2017-04-14
Task: Skip install-check for pmml

njayaram2:
JIRA: MADLIB-1078

Skip install-check for pmml when run without the '-t' option. We
can still run install-check for pmml if the '-t' option is
specified.

@iyerr3


iyerr3:
jenkins, ok to test

iyerr3:
+1

PR #116, opened 2017-04-18
Unnest 2d array

rashmi815:
Array Operations: Add function to unnest 2-D arrays into rows of 1-D arrays

JIRA:  MADLIB-1086

Function to unnest 2-D array by one level (i.e. into rows of 1-D arrays).
This is needed, for instance, in K-means, so that we can get one centroid per row for follow on operations.
- Added function to array operations
- Added an example in k-means to demonstrate usage

### DIFF
--- a/methods/array_ops/src/pg_gp/array_ops.sql_in
+++ b/methods/array_ops/src/pg_gp/array_ops.sql_in
@@ -655,7 +655,7 @@ m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `NO SQL', `');
  * @returns Set of 1-D arrays that corrspond to rows of x and an ID column.
  *
  */
-CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_unnest_2d_to_1d(x anyarray, OUT unnested_id BIGINT, OUT unnested_array_result ANYARRAY)
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_unnest_2d_to_1d(x ANYARRAY, OUT unnest_2d_to_1d_id BIGINT, OUT unnest_2d_to_1d_result ANYARRAY)
 RETURNS SETOF RECORD
 AS
 $BODY$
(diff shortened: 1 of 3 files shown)


methods/array_ops/src/pg_gp/array_ops.sql_in
```diff
@@ -636,3 +663,30 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_cum_prod(x anyarray) RETURNS anya
...
+ *        It returns a set of 1-D arrays that correspond to rows of
+ *        the input array as well as an ID column with values corresponding
+ *        to positions occupied by the 1-D arrays within the 2-D array.
+ *
+ * @param x Array x
+ * @returns Set of 1-D arrays that corrspond to rows of x and an ID column.
+ *
+ */
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_unnest_2d_to_1d(x ANYARRAY, OUT unnest_2d_to_1d_id BIGINT, OUT unnest_2d_to_1d_result ANYARRAY)
+RETURNS SETOF RECORD
```
njayaram2: Please try to have at most 80 chars in a line wherever possible, based on MADlib's coding convention.

methods/array_ops/src/pg_gp/array_ops.sql_in
```diff
@@ -636,3 +663,30 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_cum_prod(x anyarray) RETURNS anya
...
+  (
+    SELECT generate_series(array_lower($1,2),array_upper($1,2)) as c, t1.r
+    FROM
+    (
+      SELECT generate_series(array_lower($1,1),array_upper($1,1)) as r
+    ) t1
+  ) t2
+GROUP BY t2.r
+$BODY$
+LANGUAGE sql IMMUTABLE;
```
njayaram2: Having online docs with usage/example for this function would be helpful. We can have something similar to what is present in `.../incubator-madlib/methods/stemmer/src/pg_gp/porter_stemmer.sql_in`

methods/array_ops/src/pg_gp/array_ops.sql_in
iyerr3: The "unnest_2d_to_1d_id" can be an INTEGER, we won't have indices bigger than that. 
Also maybe call it just "id" or "row_id/row_num"?

iyerr3:
Jenkins, OK to test.

PR #117, opened 2017-04-18
Decision Tree: Update defaults for max_depth, num_splits

iyerr3:
Reduce the defaults for max_depth to 7 and num_splits to 20 to ensure we
don't run out of memory when initializing tree for problems with many
features or with features with many categorical values.

### DIFF
--- a/src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
+++ b/src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
@@ -223,7 +223,7 @@ SELECT {schema_madlib}.tree_train(
                                 is NULL
     'weights',              -- A Column name containing weights for
                                 each observation. Default is NULL
-    max_depth,              -- Maximum depth of any node, default is 10
+    max_depth,              -- Maximum depth of any node, default is 7
     min_split,              -- Minimum number of observations that must
                                 exist in a node for a split to be
                                 attemped, default is 20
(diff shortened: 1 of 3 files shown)


ivannovick:
does this improve memory usage?

iyerr3:
No, that's a separate JIRA: MADLIB-1057
<https://issues.apache.org/jira/browse/MADLIB-1057>. This one is just about
setting the defaults to a more reasonable value considering the data that
users have shared.

The commit is a little more than just changing two numbers since I updated
the way these defaults are set. Previously they were set in overloaded
function declaration (in SQL). Changed this to set the default in the main
function definition, eliminating redundancy.

Thanks,
Rahul

iyerr3:
Reading the commit description again, I would rephrase it as 

"Reduce the defaults for max_depth to 7 and num_splits to 20 to **minimize
the chances of running out of memory** when initializing tree for problems 
with many features or with features with many categorical values."

ivannovick:
How do we update the docs in synch with this change?
https://madlib.incubator.apache.org/docs/latest/group__grp__decision__tree.html

Is it needed?

iyerr3:
The docs/latest corresponds to latest release (1.10) and won't be updated till the next release. We also have [docs/master](http://madlib.incubator.apache.org/docs/master/), which can reflect these changes. We haven't updated those since the 1.10 release - I can update if you're looking for the changes reflected online.

fmcquillan99:
Ivan, the docs get updated when the release GA's.

PR #119, opened 2017-04-18
Multiple: Minor changes for GPDB5 and HAWQ2.2 support

orhankislal:
- Separate multi-command plpy.execute commands
- Disable some install check tests temporarily


iyerr3:
OK to test

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -314,9 +314,13 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
 						{checkg_oo})
 					UNION
 					SELECT {grp_comma} id, {weight}, parent FROM {oldupdate};
-				DROP TABLE {out_table};
-				ALTER TABLE {temp_table} RENAME TO {out_table};
-				CREATE TABLE {temp_table} AS (
+				"""
+				plpy.execute(sql.format(**locals()))
+				sql = "DROP TABLE {out_table}"
+				plpy.execute(sql.format(**locals()))
```
iyerr3: The above two lines can easily be merged to single statement (same for the ones below). 
Also avoid using locals() if there are only few variables in the format list. Explicitly adding the variables makes it easy to see their usage.

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -432,11 +433,17 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
...
-						DROP TABLE {out_table};
-						ALTER TABLE {temp_table} RENAME TO {out_table};"""
-
-				plpy.execute(sql_del.format(**locals()))
+								);"""
+					plpy.execute(sql_del.format(**locals()))
+					sql_del = "DROP TABLE {out_table}"
+					plpy.execute(sql_del.format(**locals()))
+					sql_del = "ALTER TABLE {temp_table} RENAME TO {out_table};"
+					plpy.execute(sql_del.format(**locals()))
```
iyerr3: Same as previous comment - all these `plpy.execute` can directly run the string, simplifying the code.

src/ports/postgres/modules/elastic_net/test/elastic_net_install_check.sql_in
```diff
@@ -840,27 +840,27 @@ SELECT elastic_net_train(
...
-       warmup = True, warmup_lambdas = {10, 1, 0.1}
-    $$,
-    NULL,
-    100,
-    1e-6
-);
-SELECT * FROM house_en;
-SELECT * FROM house_en_summary;
-SELECT * FROM house_en_cv;
+-- DROP TABLE if exists house_en, house_en_summary, house_en_cv;
```
iyerr3: Please add a comment here (and all other tests commented out) on why this is done. If there is a JIRA that tracks the progress of these fixes then include that here as well.

PR #120, opened 2017-04-19
DT: Assign memory only for reachable nodes

iyerr3:
JIRA: MADLIB-1057

TreeAccumulator assigns a matrix to track the statistics of rows
reaching the last layer of nodes. This matrix assumes a complete 
tree and assigns memory for all nodes. As the tree gets deeper, 
most of the nodes are unreachable, resulting in excessive wasted
memory. This commit reduces that waste by only assigning memory
for nodes that are reachable and accessing them through a lookup 
table.

### DIFF
--- a/src/modules/recursive_partitioning/decision_tree.cpp
+++ b/src/modules/recursive_partitioning/decision_tree.cpp
@@ -154,7 +154,6 @@ compute_leaf_stats_transition::run(AnyType & args){
     }
 
     if (state.empty()){
-
         // To initialize the accumulator, first find which of the leaf nodes
         // in current tree are actually reachable.
         // The lookup vector maps the leaf node index in a (fictional) complete
@@ -162,7 +161,7 @@ compute_leaf_stats_transition::run(AnyType & args){
         ColumnVector leaf_feature_indices =
             dt.feature_indices.tail(dt.feature_indices.size()/2 + 1).cast<double>();
         ColumnVector leaf_node_lookup(leaf_feature_indices.size());
-        Index n_leaves_not_finished = 0;
+        size_t n_leaves_not_finished = 0;
         for (Index i=0; i < leaf_feature_indices.size(); i++){
             if ((leaf_feature_indices(i) != dt.NODE_NON_EXISTING) &&
                     (leaf_feature_indices(i) != dt.FINISHED_LEAF)){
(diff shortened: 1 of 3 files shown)


ivannovick:
@iyerr3 how much is the impact of this change on memory reduction?  Do we have any heuristics, quantitative or qualitative estimate of reduction?

iyerr3:
@ivannovick Those are good questions. 
Short answer is it's problem (data) dependent - the memory reduction depends on how sparse the tree is.
 
I can run some experiments on public classification/regression data and give quantitative numbers on how much we would save in a typical case.

src/modules/recursive_partitioning/decision_tree.cpp
```diff
@@ -235,7 +259,14 @@ dt_apply::run(AnyType & args){
     else{
         return_code = TERMINATED;  // indicates termination due to error
     }
+    uint32_t n_valid_nodes = 0;
+    for (Index i=0; i < dt.feature_indices.size(); i++){
+        if (dt.feature_indices(i) != dt.NODE_NON_EXISTING)
+            n_valid_nodes ++;
+    }
 
+    elog(INFO, "Depth = %d, n_valid_nodes = %d",
```
orhankislal: This info and its context (the sql function that calls madlib._dt_apply) is printed to the console if the user is calling the DT from a plain psql client.

iyerr3:
@ivannovick Here's an example of how much memory can be saved. The numbers below are for a tree built for the [Poker hand dataset](https://archive.ics.uci.edu/ml/datasets/Poker+Hand), mostly using default parameters and setting `min_splits` and `min_bucket` to `1` to build a deep tree. 

Currently, memory is allocated for the maximum possible nodes, but only a fraction of that is actually used (indicated in `% usage` column). As expected, this `%` decreases as depth increases, since some nodes stop branching further. With this work, memory will be allocated at each depth only for the fraction of nodes that are actually used. 

Note, this is for a specific problem/data and results will vary for other datasets. In general, problems that are so big that they hit the 1 GB agg state limit within a couple of tree levels will not benefit for this. 

| depth | Nodes used | Max nodes (2^k - 1) | % usage | 
|-------|-------------|---------------------|---------|
| 2 | 3 | 3 | 100 | 
| 3 | 7 | 7 | 100 | 
| 4 | 11 | 15 | 73.3 | 
| 5 | 17 | 31 | 54.8 | 
| 6 | 25 | 63 | 39.7 | 
| 7 | 41 | 127 | 32.2 | 
| 8 | 73 | 255 | 28.6 | 
| 9 | 135 | 511 | 26.4 | 
| 10 | 251 | 1023 | 24.5 | 
| 11 | 447 | 2047 | 21.8  |

fmcquillan99:
That's informative, thanks for posting.

iyerr3:
ok to test

ivannovick:
cool so even in this test we see 5X reduction in memory in the case with depth of 11 right?

iyerr3:
@ivannovick you're right.

PR #123, opened 2017-04-20
MADLIB-1076. Review LICENSE file and README.md

rvs:


fmcquillan99:
Looks good, thanks for the PR.

One double check:  the LICENSE file you are proposing refers explicitly to libstemmer, useLatex and pyyaml, but does not call out other 3rd party components by name.  Is that your intention?

src/CMakeLists.txt
```diff
@@ -18,10 +18,10 @@ set(BITBUCKET_BASE_URL
     "${MADLIB_REDIRECT_PREFIX}https://bitbucket.org"
     CACHE STRING
     "Base URL for Bitbucket projects. May be overridden for testing purposes.")
-set(GITHUB_MADLIB_BASE_URL
+set(EIGEN_BASE_URL
```
iyerr3: Since this is Eigen specific link, please include the `eigen/archive` part from line 55 in the URL itself.  That way it's used specifically for that purpose.

licenses/MADlib.txt
```diff
@@ -1,10 +0,0 @@
-Portions of this software Copyright (c) 2010-2013 by EMC Corporation.  All rights reserved.
```
iyerr3: Is this file necessary?

iyerr3:
Jenkins, OK to test.

licenses/MADlib.txt
rvs: This file is actually being replaced with a symlink to ../LICENSE -- it seems to be necessary for Mac OS X packaging. I can also replace it with a copy of LICENSE, but that'll be a tad messy.

src/CMakeLists.txt
rvs: Will do. Thanks!

licenses/MADlib.txt
iyerr3: Thanks, Roman. 
Symlink would be the best option if we have to keep the file. 

Alternatively, we can change the `"${CMAKE_SOURCE_DIR}/licenses/MADlib.txt"` in `deploy/PackageMaker/CMakeLists.txt` to `"${CMAKE_SOURCE_DIR}/LICENSE"` and remove this file.

fmcquillan99:
In file ReadMe_Build.txt the lines 38-40 should be updated from

 + PostgreSQL 9.2, 9.3, 9.4
  + HAWQ 1.2, 1.3
  + Greenplum 4.2, 4.3

to 

  + PostgreSQL 9.5, 9.6
  + HAWQ 2.x
  + Greenplum 4.3.x, 5.x

PR #127, opened 2017-04-27
Build: Introduce 'dev_test' for better test coverage

njayaram2:
MADlib currently has install-check which runs a small set of test
cases to do some basic sanity check after installation. These are
expected to be light and not include complete code coverage.

This commit introduces another folder to add test cases in, very
similar to install-check. The sql files containing various test
cases must be placed in a folder named 'dev_test' inside the module
to be tested. We can have a more comprehensive set of test cases in
these sql files, that cover many more combinations of the function
parameters to use. It can be invoked similar to how install-check is
invoked, but with command 'dev-check' instead of 'install-check'
Usage example (to be run from build folder):
* src/bin/madpack -p postgres -c abc@127.0.0.1:5094/madlib dev-check
* Use dev-check with '-t' option to run on specific modules (a comma
separated list of module names following the -t option). For instance,
src/bin/madpack -p ... -c ... dev-check -t elastic_net,pca

This might be redundant with install-check around, since this does
not perform unit testing for us, but it might be useful to
have more functional test coverage at least, without overloading
install-check.
This currently has test cases only for elastic_net, very similar to
what is already present in install-check. This can be worked on and
improved if found useful.

@iyerr3 @orhankislal


njayaram2:
Closing this PR without merging, this problem requires a better solution that would cover a lot more test cases. Will be working on that sometime in the future I guess.

PR #130, opened 2017-05-04
MADLIB-1098. Corrections for MADlib naming consistency

rvs:


rvs:
Good point. Let me update the PR.

PR #133, opened 2017-05-11
Build: Add CDATA block to avoid invalid xml

iyerr3:


src/ports/postgres/modules/recursive_partitioning/test/random_forest.sql_in
```diff
@@ -9,6 +9,8 @@ CREATE TABLE dt_golf (
     class text
 ) ;
 
+INVALID SQL STATEMENT;
```
orhankislal: I assume this line was added to test error reports.

PR #135, opened 2017-05-17
Sketch: Remove per-tuple checks

iyerr3:
Some of the sketch functions have checks running for each tuple in their
aggregate. These checks include invalid transition state and invalid
types for input data. The checks are important for the functions if run
outside an aggregate context, but are a waste of cycles when called as
an agg. The checks include caql calls that were estimated to eat a large
chunk of the runtime. This work removes these checks - the average time
saved is estimated to be around 35% for datasets ranging in size from 10
million to 1 billion tuples.


fmcquillan99:
https://issues.apache.org/jira/browse/MADLIB-1104

PR #136, opened 2017-05-17
Graph: Add APSP with grouping

orhankislal:
JIRA: MADLIB-1099

- Add all pairs shortest path algorithm with grouping support.
This algorithm provides the minimum cost (sum of edge weights) paths for
every vertex pair in the graph.
- Add the get path algorithm to find the actual path for any vertex pair
(with grouping support)
- Minor changes in SSSP and graph utilities.

### DIFF
--- a/doc/mainpage.dox.in
+++ b/doc/mainpage.dox.in
@@ -125,6 +125,10 @@ complete matrix stored as a distributed table.
         @ingroup grp_datatrans
 @defgroup grp_graph Graph
 @{Contains graph algorithms. @}
+
+    @defgroup grp_apsp All Pairs Shortest Path
+    @ingroup grp_graph
+
     @defgroup grp_pagerank PageRank
     @ingroup grp_graph
 
--- a/src/ports/postgres/modules/graph/apsp.py_in
+++ b/src/ports/postgres/modules/graph/apsp.py_in
@@ -46,35 +46,12 @@ from utilities.validate_args import get_expr_type
 
 m4_changequote(`<!', `!>')
 
-# def _check_groups(tbl1, tbl2, grp_list):
-
-# 	"""
-# 	Helper function for joining tables with groups.
-# 	Args:
-# 		@param tbl1       Name of the first table
-# 		@param tbl2       Name of the second table
-# 		@param grp_list   The list of grouping columns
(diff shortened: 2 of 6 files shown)


rashmi815:
Ran a test with the following data set and found that the get_path function outputs two paths for the example I tried (tests were run using a Greenplum database 4.3.12 VM on local machine).  See below:
```sql
drop table if exists vertex_disc, edge_disc;

create table vertex_disc (id integer) distributed by (id);
create table edge_disc (src int, dest int, weight float8) distributed randomly;

insert into vertex_disc VALUES
(0),
(1),
(2),
(3),
(4),
(5),
(6),
(7),
(8),
(9),
(10)
;

insert into edge_disc VALUES
(0, 5, 1),
(1, 0, 1),
(1, 3, 1),
(2, 6, 1),
(3, 4, 1),
(3, 5, 1),
(4, 2, 1),
(8, 9, 1),
(10, 8, 1)
;

drop table if exists out_disc, out_disc_summary;
SELECT madlib.graph_apsp('vertex_disc',NULL,'edge_disc',NULL,'out_disc');

SELECT * FROM out_disc_summary;
SELECT * FROM out_disc ORDER BY src,dest;

select madlib.graph_apsp_get_path('out_disc',1,5,'out_disc_path');
select * from out_disc_path;
  path   
---------
 {1,3,5}
 {1,0,5}
(2 rows)
```

orhankislal:
@raghur1 I was trying to avoid the distinct on clause to speed up the implementation but you are right to point this out, this output is not correct. Should be fixed with my latest commit.

rashmi815:
Re-ran the test with the latest commit and it works fine now. Thanks @orhankislal . 
```sql
drop table if exists out_disc, out_disc_summary;
SELECT madlib.graph_apsp('vertex_disc',NULL,'edge_disc',NULL,'out_disc');
 graph_apsp 
------------
 
(1 row)

drop table if exists out_disc_path;
select madlib.graph_apsp_get_path('out_disc',1,5,'out_disc_path');
 graph_apsp_get_path 
---------------------
 
(1 row)


select * from out_disc_path;
  path   
---------
 {1,3,5}
(1 row)
```

PR #138, opened 2017-06-05
Summary: Add param to determine num of cols per run

iyerr3:
JIRA: MADLIB-1117

Summary used a hard-coded parameter of a maximum of 15 columns per run.
This was put in place to avoid out-of-memory errors in most cases.
This, however, limits the run time since higher number of columns can be
summarized in a single run for a simpler data set (one which leads to
smaller sketch data structures).

This commit adds a new parameter allowing users to set this limit,
while retaining the old default of 15 columns.

Closes #138

### DIFF
--- a/src/ports/postgres/modules/summary/summary.py_in
+++ b/src/ports/postgres/modules/summary/summary.py_in
@@ -7,83 +7,72 @@
 """
 import plpy
 from time import time
-from utilities.utilities import __mad_version
+
+from utilities.control import MinWarning
 from Summarizer import Summarizer
-version_wrapper = __mad_version()
-_get_vector = version_wrapper.select_vecfunc()
+
 
 def summary(schema_madlib, source_table, output_table, target_cols, grouping_cols,
-    get_distinct, get_quartiles, ntile_array, how_many_mfv, get_estimates):
+            get_distinct, get_quartiles, ntile_array, how_many_mfv,
+            get_estimates, n_cols_per_run):
     """
-        Main summary function that is called by SQL to execute summary
+        Main summary function that is called by SQL to compute summary
         statistics on a table.
 
-        @param schema_madlib        Madlib Schema namespace
-        @param source_table         Name of input table
-        @param output_table         Name of output table
-        @param target_cols          Names of specific columns for which to get summary
-        @param grouping_cols        Names of columns on which to group-by
-                                        (no summary is provided for these columns)
-        @param get_distinct         Should summary include distinct count
-        @param get_quartiles        Should summary include quartile information
-        @param ntile_array          Array for quantiles to include in summary
-                                        (each element should be in [0, 1])
-        @param how_many_mfv         How many frequent values to output?
-        @param get_estimates        Should the summmary information be estimated or exact?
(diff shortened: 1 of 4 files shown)


src/ports/postgres/modules/summary/summary.sql_in
```diff
@@ -162,31 +163,52 @@ following columns:
...
 <dd>INTEGER, default: 10. The number of most-frequent-values to compute.</dd>
+
 <dt>get_estimates (optional)</dt>
 <dd>BOOLEAN, default TRUE. If TRUE, estimated values are produced for distinct values and most frequent values. If FALSE, exact values are calculated (may take longer to run depending on data size).</dd>
+
+<dt>n_cols_per_run (optional)</dt>
+<dd>INTEGER, default: 15. The number of columns to collect summary statistics in
+one pass of the data.
+This parameter determines the number of passes through the data. For e.g.,
+with a total of 40 columns to summarize and 'n_cols_per_run = 15', there will
```
rashmi815: Needs the word "be" at the end of the line "with a total of 40 columns to summarize and 'n_cols_per_run = 15', there will "

src/ports/postgres/modules/summary/summary.sql_in
```diff
@@ -362,8 +385,9 @@ MADLIB_SCHEMA.summary
     get_quartiles           BOOLEAN,    -- Are quartiles required
     ntile_array             FLOAT8[],   -- Array of quantiles to compute
     how_many_mfv            INTEGER,    -- How many most frequent values to compute?
-    get_estimates           BOOLEAN     -- Should we produce exact or estimated
+    get_estimates           BOOLEAN,    -- Should we produce exact or estimated
                                         --      values for distinct computation
+    n_cols_per_run          INTEGER     -- Number of columns to use per run of summary
```
rashmi815: Need to add n_cols_per_run to the Python function call inside this function.

src/ports/postgres/modules/summary/summary.sql_in
rashmi815: Need to add word "be" to end of the line "with a total of 40 columns to summarize and 'n_cols_per_run = 15', there will "

iyerr3:
@rashmi815 Fixed the issues - please check if it looks good.

src/ports/postgres/modules/summary/summary.py_in
```diff
@@ -7,84 +7,73 @@
...
+        @param output_table    Name of output table
+        @param target_cols     Names of specific columns for which to get summary
+        @param grouping_cols   Names of columns on which to group-by
+                                   (no summary is provided for these columns)
+        @param get_distinct    Should summary include distinct count
+        @param get_quartiles   Should summary include quartile information
+        @param ntile_array     Array for quantiles to include in summary
+                                   (each element should be in [0, 1])
+        @param how_many_mfv    How many frequent values to output?
+        @param get_estimates   Should the summmary information be estimated or exact?
```
rashmi815: Should there be an entry here for n_cols_per_run?

PR #142, opened 2017-06-21
DT: Include NULL rows in count for termination check

iyerr3:
When the primary split feature for a node is computed, the statistics of
rows going to the true and false side don't include the rows that have
NULL value for this split feature. These "NULL" rows can only be
included in the statistics during the next pass when surrogates have
been trained. This commit ensures that in the presence of NULL rows, we
don't terminate prematurely by comparing with a lower count.

### DIFF
--- a/src/modules/recursive_partitioning/DT_impl.hpp
+++ b/src/modules/recursive_partitioning/DT_impl.hpp
@@ -453,9 +453,10 @@ DecisionTree<Container>::updatePrimarySplit(
     nonnull_split_count(trueChild(node_index)) = static_cast<double>(true_count);
     nonnull_split_count(falseChild(node_index)) = static_cast<double>(false_count);
 
-    // current node's child won't split if,
-    // 1. child is pure (responses are too similar to split further) OR
-    // 2. child is too small to split further (count < min_split)
+    // current node's each child won't split if,
+    //      1. child is pure (responses are too similar to split further)
+    //      OR
+    //      2. child is too small to split (count < min_split)
     bool children_wont_split = ((isChildPure(true_stats) ||
                                     true_count < min_split) &&
                                 (isChildPure(false_stats) ||
@@ -485,20 +486,23 @@ DecisionTree<Container>::expand(const Accumulator &state,
             Index stats_i = static_cast<Index>(state.stats_lookup(i));
             assert(stats_i >= 0);
 
+            // 1. Update predictions if necessary
             if (statCount(predictions.row(current)) !=
                     statCount(state.node_stats.row(stats_i))){
                 // Predictions for each node is set by its parent using stats
-                // recorded while training parent node. These stats do not include
-                // rows that had a NULL value for the primary split feature.
-                // The NULL count is included in the 'node_stats' while training
-                // current node. Further, presence of NULL rows indicate that
-                // stats used for deciding 'children_wont_split' are inaccurate.
-                // Hence avoid using the flag to decide termination.
+                // recorded while training parent node. These stats do not
+                // include rows that had a NULL value for the primary split
+                // feature. The NULL count is included in 'node_stats' while
+                // training current node.
                 predictions.row(current) = state.node_stats.row(stats_i);
+
+                // Presence of NULL rows indicate that stats used for deciding
+                // 'children_wont_split' are inaccurate. Hence avoid using the
+                // flag to decide termination.
                 children_wont_split = false;
             }
 
-            // 2. Compute the best feature to split current node by
+            // 2. Compute the best feature to split current node
 
             // if a leaf node exists, compute the gain in impurity for each split
             // pick split  with maximum gain and update node with split value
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/84/

asfgit:
https://builds.apache.org/job/madlib-pr-build/86/

src/modules/recursive_partitioning/DT_impl.hpp
```diff
@@ -446,21 +446,20 @@ DecisionTree<Container>::updatePrimarySplit(
...
-    bool children_wont_split = (isChildPure(true_stats) &&
-                                isChildPure(false_stats) &&
-                                true_count < min_split &&
-                                false_count < min_split
-                                );
+    nonnull_split_count(trueChild(node_index)) = static_cast<double>(true_count);
+    nonnull_split_count(falseChild(node_index)) = static_cast<double>(false_count);
+
+    // current node's child won't split if,
+    // 1. child is pure (responses are too similar to split further) OR
```
orhankislal: Why did we change the logic on children_wont_split?

src/modules/recursive_partitioning/DT_impl.hpp
```diff
@@ -486,8 +485,18 @@ DecisionTree<Container>::expand(const Accumulator &state,
...
+                // Predictions for each node is set by its parent using stats
+                // recorded while training parent node. These stats do not include
+                // rows that had a NULL value for the primary split feature.
+                // The NULL count is included in the 'node_stats' while training
+                // current node. Further, presence of NULL rows indicate that
+                // stats used for deciding 'children_wont_split' are inaccurate.
+                // Hence avoid using the flag to decide termination.
+                predictions.row(current) = state.node_stats.row(stats_i);
+                children_wont_split = false;
+            }
```
orhankislal: Should we move the sections 2 and 3 under the else clause of this if check (except the `children_not_allocated` check at line 559)? Since `children_wont_split` is set to false, the calculation at line 564 should be redundant.

src/modules/recursive_partitioning/DT_impl.hpp
iyerr3: I've added a new commit with more explanation. The short answer is that the previous logic was incorrect and resulting in longer trees.

src/modules/recursive_partitioning/DT_impl.hpp
iyerr3: The if statement is basically checking if a NULL row is present in the `current` node and if yes, then the predictions for that node is updated. I've added an explanation in the comments for both statements on why they're needed. If the explanation is not clear then please add more details on what would help you understand.

iyerr3:
I need more info on the RF docs discrepancy. 
- Which example is giving the lower than 100% training accuracy? 
- How do the trees look? 
- Can we replicate in decision tree since that is easier to debug? 

On its own, less than 100% accuracy is not wrong, but if the tree is not as long as it should be (i.e. prematurely terminating) then a problem has been introduced here.

asfgit:
https://builds.apache.org/job/madlib-pr-build/88/

src/modules/recursive_partitioning/DT_impl.hpp
orhankislal: Let me try to rephrase my question:
Assume the if check at lines 490-491 is `true`. `children_wont_split` will be set to `false`. This variable is used with an & operator at line 568. That means the second boolean value is irrelevant, the result will always be `false`. In this case, do we need the 2 double for loops from lines 516-547? 

The compiler might calculate the backwards slice of this instruction to find its optimal location but I don't think we should rely on that.

iyerr3: - `children_wont_split` is **one** of the factors that determines if training should stop after current iteration. `children_wont_split=true` implies training stops; `children_wont_split=false` implies other flags determine termination. 
- The lines 516-547 are finding the best feature to split on and are necessary - independent of `children_wont_split` and independent of the result of line 490. 

I could exchange sections 1 and 2 since they're independent, if that helps in reading the code.

iyerr3:
When I ran the RF example thrice - twice I got 100% and once I got 13/14 (~93%). I guess there's some randomness there (which is expected).

asfgit:
https://builds.apache.org/job/madlib-pr-build/89/

src/modules/recursive_partitioning/DT_impl.hpp
orhankislal: Thanks, I agree that this makes more sense.

src/modules/recursive_partitioning/DT_impl.hpp
orhankislal: No need to exchange anything. After reading through the `updatePrimarySplit` function, I understand why it is needed even if its boolean output doesn't affect the variable.

orhankislal:
DT output seems correct, it seems RT randomness was the cause. LGTM +1

PR #148, opened 2017-07-08
Graph: Update Python code to follow PEP-8

iyerr3:
- Changed indentation to use spaces instead of tabs
- Updated to PEP-8 guidelines
- Updated to follow style guide convention
- Refactored few functions to clean code and design

### DIFF
--- a/src/ports/postgres/modules/graph/graph_utils.py_in
+++ b/src/ports/postgres/modules/graph/graph_utils.py_in
@@ -27,104 +27,103 @@
 @namespace graph
 """
 
-import plpy
-from utilities.control import MinWarning
 from utilities.utilities import _assert
-from utilities.utilities import extract_keyvalue_params
-from utilities.utilities import unique_string
 from utilities.validate_args import get_cols
 from utilities.validate_args import unquote_ident
 from utilities.validate_args import table_exists
 from utilities.validate_args import columns_exist_in_table
 from utilities.validate_args import table_is_empty
 
+
 def _check_groups(tbl1, tbl2, grp_list):
+    """
+    Helper function for joining tables with groups.
+    Args:
+            @param tbl1       Name of the first table
+            @param tbl2       Name of the second table
+            @param grp_list   The list of grouping columns
+    """
 
-	"""
-	Helper function for joining tables with groups.
-	Args:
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/108/

src/ports/postgres/modules/graph/graph_utils.py_in
```diff
@@ -27,104 +27,103 @@
...
+    ----------------------------------------------------------------------------
+                                USAGE
+    ----------------------------------------------------------------------------
+     SELECT {schema_madlib}.{func_name}(
+        vertex_table  TEXT, -- Name of the table that contains the vertex data.
+        vertex_id     TEXT, -- Name of the column containing the vertex ids.
+        edge_table    TEXT, -- Name of the table that contains the edge data.
+        edge_args     TEXT{comma} -- A comma-delimited string containing multiple
+                            -- named arguments of the form "name=value".
+        {other_text}
```
njayaram2: `other_text` is also a string that is passed to this function. Shouldn't that string also be indented with a tab's space? For instance, this is how help on `sssp` looks like now:
```
                                    graph_sssp
----------------------------------------------------------------------------------
                                                                                 
 Given a graph and a source vertex, single source shortest path (SSSP)           
 algorithm finds a path for every vertex such that the sum of the                
 weights of its constituent edges is minimized.                                  
                                                                                 
                                                                                 
     ----------------------------------------------------------------------------
                                 USAGE                                           
     ----------------------------------------------------------------------------
      SELECT madlib.graph_sssp(                                                  
         vertex_table  TEXT, -- Name of the table that contains the vertex data. 
         vertex_id     TEXT, -- Name of the column containing the vertex ids.    
         edge_table    TEXT, -- Name of the table that contains the edge data.   
         edge_args     TEXT, -- A comma-delimited string containing multiple     
                             -- named arguments of the form "name=value".       
         source_vertex INT,  -- The source vertex id for the algorithm to start. 
     out_table     TEXT, -- Name of the table to store the result of SSSP.    
     grouping_cols TEXT  -- The list of grouping columns.                        
     );                                                                          
```
the description of `out_table` and `grouping_cols` comes from the `other_text` param.

PR #194, opened 2017-11-06
Logregr: Add input validation for dep/indep variables

orhankislal:


asfgit:
https://builds.apache.org/job/madlib-pr-build/267/

asfgit:
https://builds.apache.org/job/madlib-pr-build/270/

src/ports/postgres/modules/regress/logistic.py_in
```diff
@@ -158,12 +159,14 @@ def __logregr_validate_args(schema_madlib, tbl_source, tbl_output, dep_col,
     if not dep_col or dep_col.strip().lower() in ('null', ''):
         plpy.error("Logregr error: Invalid dependent column name!")
 
-    # if not columns_exist_in_table(tbl_source, [dep_col]):
-    #     plpy.error("Logregr error: Dependent column does not exist!")
+    if not is_var_valid(tbl_source, dep_col):
+        plpy.error("Logregr error: Dependent variable is not valid!")
```
kaknikhil: should we print the name of the dependent variable as well and if possible maybe say why it's not valid i.e column does not exist in the table tbl_source ?
same for the other is_var_valid calls.

orhankislal: Since the variable can be an expression and not a column, I wanted to avoid printing the whole expression and making the error message long and confusing. We can easily add it if you feel that would be more useful.

orhankislal:
I have decided to close this pull request since the default error given by the database is more descriptive.

PR #195, opened 2017-11-07
 Feature: Add grouping support to HITS 

kaknikhil:
JIRA: MADLIB-1151

Changes to support grouping column in HITS. Update queries to
use group by and other necessary sql constructs.


asfgit:
https://builds.apache.org/job/madlib-pr-build/268/

njayaram2:
Jenkins ok to test

asfgit:
https://builds.apache.org/job/madlib-pr-build/269/

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/272/

src/ports/postgres/modules/graph/graph_utils.py_in
```diff
@@ -109,6 +110,85 @@ def validate_graph_coding(vertex_table, vertex_id, edge_table, edge_params,
 
     return None
 
+def validate_params_for_centrality_measures(schema_madlib, func_name,
```
orhankislal: This function name is a bit confusing since it isn't used by the centrality measures functions from `measures.py_in`

src/ports/postgres/modules/graph/graph_utils.py_in
```diff
@@ -109,6 +110,85 @@ def validate_graph_coding(vertex_table, vertex_id, edge_table, edge_params,
...
+            """{0}: Invalid max_iter value ({1}), must be a positive integer.""".
+            format(func_name, max_iter))
+    if grouping_cols_list:
+        # validate the grouping columns. We currently only support grouping_cols
+        # to be column names in the edge_table, and not expressions!
+        _assert(columns_exist_in_table(edge_table, grouping_cols_list, schema_madlib),
+                "{0} error: One or more grouping columns specified do not exist!".
+                format(func_name))
+
+def update_output_grouping_tables_for_centrality_measures(temp_summary_table,
```
orhankislal: Maybe a shorter name will help us keeping the lines under 80 characters.

src/ports/postgres/modules/utilities/utilities.py_in
```diff
@@ -709,6 +709,17 @@ def _check_groups(tbl1, tbl2, grp_list):
     return ' AND '.join([" {tbl1}.{i} = {tbl2}.{i} ".format(**locals())
                          for i in grp_list])
 
+def get_ignore_groups(first_table, second_table, grouping_cols_list):
```
iyerr3: This function (and the `_grp_from_table`) are difficult to understand just from the names (for eg. what does `first_table` and `second_table` impy?). Further, these are returning specific SQL expressions - is there value in putting these in a general `utilities` module? If yes, then I would suggest making the use case more general. 
The `_grp_from_table` especially seems like extra work for a one-liner, without getting benefits of abstraction.

src/ports/postgres/modules/utilities/validate_args.py_in
```diff
@@ -262,6 +262,13 @@ def get_first_schema(table_name):
     return None
 # -------------------------------------------------------------------------
 
+def drop_tables(table_list):
+    """
+        Drop tables specified in table_list.
+    """
+    drop_str = ', '.join([table for table in table_list])
```
iyerr3: Why not just `', '.join(table_list)`?

iyerr3:
Not sure what the `*_for_centrality_measures` names mean - are those function only used in centrality measures?

kaknikhil:
@iyerr3  @orhankislal All the `*_for_centrality_measures` functions are currently only used in `pagerank` and `hits`, which are centrality measures. They can potentially be used in future centrality measure modules, hence the generic name and not something like `*hits_pagerank*`. 

Any suggestions on what is a better naming convention?

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/280/

kaknikhil:
@orhankislal @iyerr3 we have refactored the functions grp_frm_table and get_ignore_groups. Let us know if it looks good ?

src/ports/postgres/modules/utilities/utilities.py_in
```diff
@@ -709,16 +709,35 @@ def _check_groups(tbl1, tbl2, grp_list):
     return ' AND '.join([" {tbl1}.{i} = {tbl2}.{i} ".format(**locals())
                          for i in grp_list])
 
-
-def _grp_from_table(tbl, grp_list):
-    """
-    Helper function for selecting grouping columns of a table
+def get_filtered_cols_subquery_str(include_from_table, exclude_from_table,
```
jingyimei: what is our consideration to define function name not start with underscore(weak internal use) while the previous one is a weak internal use?

src/ports/postgres/modules/graph/hits.sql_in
```diff
@@ -164,10 +168,20 @@ INSERT INTO edge VALUES
...
+(1, 3, 2),
+(2, 3, 2),
+(3, 0, 2),
+(4, 0, 2),
+(5, 6, 2),
+(6, 3, 2);
 </pre>
 
--# Compute the HITS scores:
+-# Running HITS with default values for optional parameters:
```
jingyimei: In those examples without grouping, duplicate edges from different groups are calculated twice. We discussed this before and just wondering if we can add some explanation here(may not be necessary tho).

src/ports/postgres/modules/graph/graph_utils.py_in
jingyimei: +1

src/ports/postgres/modules/graph/hits.py_in
```diff
@@ -95,234 +109,391 @@ def hits(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
+        # Set initial authority_norm and hub_norm as 1, so that later the final
+        # norm should be positive number
+        authority_init_value = 1.0
+        hub_init_value = 1.0
+
+        subquery1 = unique_string(desp='subquery1')
+
+        distinct_grp_table = ''
+        select_grouping_cols_comma = ''
+        select_subquery1_grouping_cols_comma = ''
```
jingyimei: I understood the incentive to rename those variables to make them more meaningful, however, seems this doesn't help simplify/reduce code but extends the length of many lines. When I read the code I still feel hard to keep track of them and understand the query. In this case I feel a shorter variable name is even simpler to read.

src/ports/postgres/modules/graph/hits.sql_in
kaknikhil: @jingyimei good point. we checked in a commit to remove the extra user_id `2` from the edge table when hits is called without the grouping column.

src/ports/postgres/modules/utilities/utilities.py_in
kaknikhil: we removed the _ because get_filtered_cols_subquery_str is not for internal use, it is called by pagerank and the python files.

src/ports/postgres/modules/graph/hits.py_in
kaknikhil: While this may not be perfect, we felt that it is slightly better than having short variable names that do not convey much meaning at all.

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/281/

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/283/

src/ports/postgres/modules/graph/hits.py_in
iyerr3: My 2 cents: 
1. Descriptive names are definitely preferred. Here, however, it's not clear how `grouping_cols_comma` differs from `select_grouping_cols_comma` which differs from `select_subquery1_grouping_cols_comma` just by looking at the values.  I prefer for the variable name should indicate *what* the value represents rather than *where* it's used. 
2. My preference is to have an else clause that defines these empty variables. That makes it clear that the variables can have two possible values: one for grouping and the empty for non-grouping.

src/ports/postgres/modules/utilities/utilities.py_in
```diff
@@ -709,16 +709,35 @@ def _check_groups(tbl1, tbl2, grp_list):
...
+                                   filter_cols_list):
+    """
+    This function returns a subquery string with columns in the filter_cols_list
```
iyerr3: I'm confused with this description string. My understanding is that the function doesn't really filter the columns - it filters the values for the columns i.e. it returns a subquery string that would filter values in exclude_from_table. If you're interested in filtering columns, then wouldn't querying the catalog and then doing a set difference be the better option?

PR #197, opened 2017-11-13
Fix madlib version parsing for upgrade

orhankislal:


src/madpack/upgrade_util.py
```diff
@@ -142,11 +142,11 @@ def _load(self):
         """
 
         # _mad_dbrev = 1.9.1
-        if self._mad_dbrev.split('.') < '1.10.0'.split('.'):
+        if map(int,self._mad_dbrev.split('.')) < map(int,'1.10.0'.split('.')):
```
iyerr3: I would recommend using the `_is_rev_gte` and `_get_rev_num` functions as used in `madpack.py` (eg. https://github.com/apache/madlib/blob/aed4d9706e59c781f4a83a023358b70b65b84085/src/madpack/madpack.py#L658)

iyerr3: When we refactor `madpack.py`, we should move these function to a common utility module and make them non-private. IMO, it's OK to use them as is till that happens.

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/273/

src/madpack/upgrade_util.py
orhankislal: Importing those files from madpack.py creates a dependency circle.

iyerr3: Multiple ways to work around that. Ideally, move common functions to a separate module. But easier solution would be to do `import madpack` instead of `from madpack import ...` and then use `madpack._is_rev_gte`.

orhankislal: I was thinking of the first option as you suggested. I am not sure about your second suggestion. I tried all 4 of the combinations and couldn't get it working.

iyerr3: The change would have to go in both modules i.e. in `madpack.py` the import statement will have to be `import upgrade_util` instead of the `from upgrade_util import ...`. 
Alternatively, you could move the functions to `upgrade_util.py`, thus eliminating the circular imports.

asfgit:
FAILURE

asfgit:
https://builds.apache.org/job/madlib-pr-build/276/

orhankislal:
jenkins ok to test

asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/277/

PR #201, opened 2017-11-15
Allow array feature with more than 1664 entries

iyerr3:
JIRA: MADLIB-1173

The tree_predict function concatenates cat_feature_str and
con_feature_str in summary table to obtain the feature string. This
contains individual elements of any array feature. The concatenated
string is used in a SELECT operation for validation, which limits the
number of target entries to 1664. To allow arrays with more than 1664
entries, the validation has been udpated to use the original feature
string, which contains the array feature name instead of its indexed
elements.

Closes #201


asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/279/

PR #203, opened 2017-11-17
Build: Create single binary for all PG10 versions

iyerr3:
JIRA: MADLIB-1179

Postgresql starting 10.0 is switching to semantic versioning (see
https://www.postgresql.org/support/versioning). Hence, we can build a
single MADlib binary that supports all 10.x PG versions.


asfgit:
SUCCESS

asfgit:
https://builds.apache.org/job/madlib-pr-build/284/

PR #210, opened 2017-12-06
Regularized Regression: Change cross validation stats

njayaram2:
JIRA:MADLIB-1169

Cross Validation seems to be supported by Elastic Net, SVM, and
Decision Trees. If a module is run with cross validation optimization
params, the output table corresponding to it displays `mean` and
`std` of the negative loss error for each permutation of the CV
params.
- This commit changes column names: `mean`->`mean_neg_loss` and
`std`->`std_neg_loss`.
- CV now uses negative Root Mean Squared Error, instead
of the negative Mean Squared Error.
- Update Elastic Net user docs to reflect these changes.

Closes #210


asfgit:
https://builds.apache.org/job/madlib-pr-build/296/

PR #213, opened 2017-12-11
KNN: Move online help to python layer

orhankislal:
Additional Author: Nikhil Kak <nkak@pivotal.io>

- Remove the dependency on the client message level for knn online help.

PR #215, opened 2017-12-12
Modify knn help funtion for easier use : JIRA-1187

jingyimei:
This commit does the following:
1. instead of a `RAISE NOTICE` in udf definition, implement it in a
plpython so the help message can always get printed out.
2. add 'example' also as a valid input argument, and make all the input
arguments ignore upper and lower cases
3. Add examples in help funtion while there was no examples before


asfgit:
https://builds.apache.org/job/madlib-pr-build/306/

iyerr3:
Is there overlap between this and #213? Maybe consolidate to a single PR?

PR #219, opened 2017-12-21
Multiple: Hard-wire values for construct_array calls

iyerr3:
JIRA: MADLIB-1185

Original investigation and RCA performed by
  Nikhil Kak <nkak@pivotal.io> and
  Orhan Kislal <okislal@pivotal.io>

Multiple modules called get_typlenbyvalalign in the constructor of a
struct, which led to querying the catalog during dlopen. This is frowned
down upon [1] and the problem got exposed in PG10. This commit provides
a temporary solution by hard-wiring the necessary values. This is OK in
these scenarios, since the types used are INT4, INT8 and FLOAT8, all
with fixed, stable storage and alignment patterns. Ideal solution to the
problem is to use MADlib's Allocator::allocateArray instead of the
construct_array functions.

[1] https://www.postgresql.org/message-id/96420364a3d055172776752a1de80714%40smtp.hushmail.com

Closes #219

### DIFF
--- a/src/modules/crf/viterbi.cpp
+++ b/src/modules/crf/viterbi.cpp
@@ -18,22 +18,8 @@ namespace modules {
 namespace crf {
 
 using namespace dbal::eigen_integration;
-using madlib::dbconnector::postgres::madlib_get_typlenbyvalalign;
 using madlib::dbconnector::postgres::madlib_construct_array;
 
-typedef struct __type_info{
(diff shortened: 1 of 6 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/316/

orhankislal:
LGTM +1

PR #220, opened 2017-12-22
  Add more stats to summary function

jingyimei:
This PR added the following statistics to madlib.summary():
    positive values
    negative values
    zero values
    95% confidence intervals on mean
User docs is updated due to adding new fields.

Besides, we rename  'row_count' to 'num_col_summarized' in summary() return result to eliminate confusion from another `row_count` in user_defined_summary_table.


src/ports/postgres/modules/summary/Summarizer.py_in
```diff
@@ -199,6 +200,22 @@ class Summarizer:
...
+                                           then 1 else 0 end)".format(c['attname'])
+                                          if c['typname'] in numeric_types
+                                          else 'NULL' for c in cols])
+
+        args["negative_columns"] = ','.join(["sum(case when {0} < 0 \
+                                           then 1 else 0 end)".format(c['attname'])
+                                          if c['typname'] in numeric_types
+                                          else 'NULL' for c in cols])
+
+        args["zero_columns"] = ','.join(["sum(case when {0} = 0 \
```
fmcquillan99: should we check equality to 0 exactly like this, or abs value < SMALL ?

asfgit:
https://builds.apache.org/job/madlib-pr-build/317/

src/ports/postgres/modules/summary/Summarizer.py_in
jingyimei: If we are using float, maybe this will be more accurate.

orhankislal: In graph algorithms such as SSSP and APSP, we used `EPSILON = 0.000001` for float comparisons.

fmcquillan99: Orhan when you say "comparison" between FLOATs, then yes, we do need to consider rounding.    

But if we are checking *explicit* equality to 0 in summary() function, then we should not worry about rounding, it seems to me.  So I am changing my opinion here, and suggest that this PR that does checking = 0 is fine.  We can put it in the user docs that this is what we are doing.   

Others please comment if you think this is an OK approach.

iyerr3:
This can be closed since merged by d025bb4609baeb7c7a1d136590780a8fafdee208.

PR #221, opened 2017-12-26
PGXN: Add newline between files during concat

iyerr3:
PGXN: Add newline between files during concat

JIRA: MADLIB-1194

PGXN requires a single SQL file during installation. This is created by
concatenating all our sql_in files together. This was earlier performed
with a simple `cat` command, but that led to an error due to a missing
newline at EOF (MADLIB-1191). This commit avoids such an issue by
explicitly adding a newline during the concatenation. This is performed
by using `awk` instead of `cat`.

Closes #221


asfgit:
https://builds.apache.org/job/madlib-pr-build/318/

PR #224, opened 2018-01-12
1.13 Upgrade and MLP IC fix

orhankislal:
JIRA: MADLIB-1197

Additional Author: Nandish Jayaram <njayaram@apache.org>

- 1.13 Upgrade does not drop the kNN help functions even though their
return types are changed. This commit adds the missing functions to the
changelist and alters the upgrade_util.py_in so that functions without
arguments can be dropped.

- Some assert thresholds are too strict for MLP in IC. This commit relaxes
those thresholds.

Closes #224


asfgit:
https://builds.apache.org/job/madlib-pr-build/323/

PR #231, opened 2018-02-06
RF: Output non-negative importance values

iyerr3:
Variable importance is computed in RF as the difference in prediction
accuracy between original data and permuted data from out-of-bag
samples (OOB). Permuted data is defined as each variable resampled from
its own distribution. This value can end up being negative if the number
of levels for a variable is small and is unbalanced, as the
redistribution doesn't change the data much. This commit shifts all the
importance values if some of them are negative to ensure that the lowest
importance value is 0.

Closes #231

### DIFF
--- a/src/ports/postgres/modules/recursive_partitioning/random_forest.py_in
+++ b/src/ports/postgres/modules/recursive_partitioning/random_forest.py_in
@@ -471,8 +471,8 @@ def forest_train(
                             1 AS gid,
                             {dep} AS dep,
                             {dep} AS oob_prediction,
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/336/

fmcquillan99:
Does this mean, then, that all var importance values are >= 0 now, and that the largest positive value corresponds to the most "important" variable?

Also, what is the range of possible values for variable performance?

iyerr3:
This change ensures that all variable importance values are positive. The
remaining properties remain as is: i.e. the feature with max value is most
important and the values are not normalized.

PR #239, opened 2018-03-07
Balance Sample: Add support for grouping

iyerr3:
JIRA: MADLIB-1168

This commit adds grouping support for balanced sampling. 
Grouping is implemented as a loop over the existing logic, 
with the sampling for each group run independently. 

Closes #239

### DIFF
--- a/src/ports/postgres/modules/sample/balance_sample.py_in
+++ b/src/ports/postgres/modules/sample/balance_sample.py_in
@@ -562,10 +562,10 @@ def balance_sample(schema_madlib, source_table, output_table, class_col,
         bool_to_text_patched_col = explicit_bool_to_text(source_table,
                                                          [class_col],
                                                          schema_madlib)[0]
-        distinct_levels = collate_plpy_result(
-            plpy.execute("SELECT DISTINCT ({0})::TEXT as levels FROM {1} ".
-                         format(bool_to_text_patched_col,
-                                source_table)))['levels']
+        distinct_sql = ("SELECT DISTINCT ({0})::TEXT as levels FROM {1} ".
+                        format(bool_to_text_patched_col,
+                               source_table))
+        distinct_levels = collate_plpy_result(plpy.execute(distinct_sql))['levels']
         if not parsed_class_sizes:
             sampling_strategy_str = _validate_and_get_sampling_strategy(
                 class_sizes, output_table_size)
--- a/src/ports/postgres/modules/sample/balance_sample.py_in
+++ b/src/ports/postgres/modules/sample/balance_sample.py_in
@@ -87,7 +87,7 @@ def _get_level_frequency_distribution(source_table, class_col,
             array_agg(classes) as classes,
             array_agg(class_count) as class_count
         FROM(
-            -- for each group, class combination present in source table
+            -- for each group and class combination present in source table
             -- get the count of rows for that combination
             SELECT
                 {array_grp_by_cols_comma}
@@ -106,6 +106,8 @@ def _get_level_frequency_distribution(source_table, class_col,
 
     actual_level_counts = {}
     for each_row in query_result:
+        # group_values is a list for each row; convert it to a tuple to use as
+        # key in a dictionary
         grp = tuple(each_row['group_values']) if is_grouping else None
         grp_levels, grp_counts = each_row['classes'], each_row['class_count']
         actual_level_counts[grp] = dict(zip(grp_levels, grp_counts))
@@ -186,43 +188,52 @@ def _choose_strategy(actual_count, desired_count):
 
 
 def _get_desired_target_level_counts(desired_level_counts, actual_level_counts,
-                                     output_table_size, grp_values=None):
+                                     output_table_size, n_grp_values=1):
     """ Returns the target level counts for each class, based on the user
         defined class_size string. The strategy of (under)oversampling for each
         class level is chosen based on the desired number of counts for a level,
         and the actual number of counts already present in the input table.
 
     """
-    if not grp_values:
-        grp_values = [True]
     target_level_counts = {}
     try:
         for each_level, desired_count in desired_level_counts.items():
+            # actual_level_counts are already group specific, whereas
+            # desired_level_counts are for all groups
+            per_grp_desired_count = math.ceil(float(desired_count) / n_grp_values)
             sample_strategy = _choose_strategy(actual_level_counts[each_level],
-                                               desired_count)
-            target_level_counts[each_level] = (desired_count, sample_strategy)
+                                               per_grp_desired_count)
+            target_level_counts[each_level] = (per_grp_desired_count,
+                                               sample_strategy)
     except KeyError:
+        grp_error = "for each group " if n_grp_values > 1 else ""
         plpy.error("Balance sample: Desired class level ({0}) not present in the "
-                   " data".format(each_level))
+                   "data {1}".format(each_level, grp_error))
+
+    # desired levels can only have a subset of all levels. For the remaining
+    # levels in actual_level_counts, compute the desired counts
     remaining_levels = (set(actual_level_counts.keys()) -
                         set(desired_level_counts.keys()))
-    #   if 'output_table_size' = NULL, unspecified level counts remain as is
-    #   if 'output_table_size' = <Integer>, divide remaining row count
-    #                             uniformly among unspecified level counts
+    #   if 'output_table_size' = NULL, remaining level counts remain as is
+    #   if 'output_table_size' = <Integer>, divide remaining count
+    #                             uniformly among reamining levels
     if output_table_size:
-        # Uniformly distribute across the remaining class levels
-        remaining_rows = output_table_size - sum(desired_level_counts.values())
+        # output_table_size is for the whole table and should be split evenly
+        # between the groups
+        remaining_rows = math.ceil(float(output_table_size -
+                                         sum(desired_level_counts.values())) /
+                                   n_grp_values)
         if remaining_rows > 0:
+            # Uniformly distribute the remaining class levels
             rows_per_level = math.ceil(float(remaining_rows) /
                                        len(remaining_levels))
             for each_level in remaining_levels:
                 sample_strategy = _choose_strategy(
                     actual_level_counts[each_level], rows_per_level)
-                target_level_counts[each_level] = (rows_per_level,
-                                                   sample_strategy)
+                target_level_counts[each_level] = (rows_per_level, sample_strategy)
     else:
         # When output_table_size is unspecified, rows from the input table
-        # are sampled as is for remaining class levels. This is same as the
+        # are sampled as is for remaining class levels. This is called as the
         # NOSAMPLE strategy.
         for each_level in remaining_levels:
             target_level_counts[each_level] = (actual_level_counts[each_level],
@@ -233,7 +244,8 @@ def _get_desired_target_level_counts(desired_level_counts, actual_level_counts,
 
 def _get_supported_target_level_counts(sampling_strategy_str,
                                        actual_level_counts,
-                                       output_table_size):
+                                       output_table_size,
+                                       n_grp_values=1):
     """ Returns the target level counts for all levels when the class_size param
         is one of [uniform, undersample, oversample]. The strategy of
         (under)oversampling for a specific class level is chosen based on the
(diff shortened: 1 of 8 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/361/

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -149,8 +149,10 @@ non-stratified, that is, the whole table is treated as a
 single group.
 
 @note
-Current implementation does not support grouping_cols.  It
-will be added in an upcoming release.
+The 'output_table_size' and the 'class_sizes' are defined for the whole table.
+When grouping is used, these parameters are split evenly for each group.
+Further, if a specific class value is specified in the 'class_sizes' parameter,
+that particular class value should be present in each group.
 </dd>
```
fmcquillan99: Please change/add:
"...that particular class value should be present in each group."
to 
"...that particular class value should be present in each group.  If not, an error will be thrown."

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +544,90 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+<pre class="syntax">
+DROP TABLE IF EXISTS output_table;
+SELECT madlib.balance_sample(
+    'flags',          -- Source table
+    'output_table',   -- Output table
+    'mainhue',        -- Class column
+    NULL,             -- Uniform
+    NULL,             -- Output table size
+    'zone'            -- No grouping
+);
```
fmcquillan99: I think you mean "Group by zone" or something like that, not "No grouping" .

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +544,90 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+<pre class="syntax">
+DROP TABLE IF EXISTS output_table;
+SELECT madlib.balance_sample(
+    'flags',          -- Source table
+    'output_table',   -- Output table
+    'mainhue',        -- Class column
+    'blue=2',         -- Specified class value size. Rest of the values are sampled as is.
+    NULL,             -- Output table size
+    'zone'            -- No grouping
+);
```
fmcquillan99: I think you mean "Group by zone" or something like that, not "No grouping" .

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +544,90 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+            12 | 13 | Mexico      |        1 |    4 | 1973 |         77 |        2 |       4 | green
+            10 | 13 | Mexico      |        1 |    4 | 1973 |         77 |        2 |       4 | green
+            11 | 13 | Mexico      |        1 |    4 | 1973 |         77 |        2 |       4 | green
+             1 | 19 | UK          |        3 |    4 |  245 |         56 |        1 |       3 | red
+             3 |  5 | Canada      |        1 |    4 | 9976 |         24 |        1 |       2 | red
+             2 | 15 | Portugal    |        3 |    4 |   92 |         10 |        6 |       5 | red
+             8 | 20 | USA         |        1 |    4 | 9363 |        231 |        1 |       3 | white
+             7 | 20 | USA         |        1 |    4 | 9363 |        231 |        1 |       3 | white
+             9 | 10 | Ireland     |        3 |    4 |   70 |          3 |        1 |       3 | white
+(23 rows)
```
fmcquillan99: For zone 4 there are 12 output rows sampled from 9 input rows.  I am assuming this is just rounding due to small numbers in the example?  i.e., could have been 8 rows or 12 rows in the output table for uniform sampling for 4 groups (blue, green, red, white) but we went with 12? 

Which is fine, just confirming.

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +544,90 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+-# Grouping can be used with class size specification as well. Note below that
+'blue=<Integer>' is the only valid class value since 'blue' is the only class
+value that is present in each group.
+<pre class="syntax">
+DROP TABLE IF EXISTS output_table;
+SELECT madlib.balance_sample(
+    'flags',          -- Source table
+    'output_table',   -- Output table
+    'mainhue',        -- Class column
+    'blue=2',         -- Specified class value size. Rest of the values are sampled as is.
```
fmcquillan99: Do you mean 
“Rest of the values are left as is.”
since they are not actually sampled. ?

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +544,90 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+             1 |  9 | Guatemala   |        1 |    4 |  109 |          8 |        2 |       2 | blue
+             4 | 11 | Jamaica     |        1 |    4 |   11 |          2 |        1 |       3 | green
+             5 | 13 | Mexico      |        1 |    4 | 1973 |         77 |        2 |       4 | green
+             8 | 19 | UK          |        3 |    4 |  245 |         56 |        1 |       3 | red
+             2 |  5 | Canada      |        1 |    4 | 9976 |         24 |        1 |       2 | red
+             7 | 16 | Spain       |        3 |    4 |  505 |         38 |        2 |       2 | red
+             6 | 15 | Portugal    |        3 |    4 |   92 |         10 |        6 |       5 | red
+             9 | 20 | USA         |        1 |    4 | 9363 |        231 |        1 |       3 | white
+             3 | 10 | Ireland     |        3 |    4 |   70 |          3 |        1 |       3 | white
+(19 rows)
```
fmcquillan99: This example asks for `blue=2` for class_size but the output has blue=1 for all groups.  Is this a bug or a cut and paste error in the example?

src/ports/postgres/modules/sample/balance_sample.sql_in
iyerr3: Right. We're always using ceil instead of round and hence will provide more than the expected rows in cases where division is not exact.

src/ports/postgres/modules/sample/balance_sample.sql_in
iyerr3: This is because the blue=2 is spilt between the groups. I guess I could change the example to use blue=8 to ensure 2 blues for each group.

fmcquillan99: I forgot that blue=2 means for the whole table, so my misunderstanding, but perhaps others may be confused too.

So yes, please put blue=8 in the example and remind the user that this applies to the whole table so that each group will get 2 blues each.

src/ports/postgres/modules/sample/balance_sample.py_in
```diff
@@ -468,81 +544,107 @@ def balance_sample(schema_madlib, source_table, output_table, class_col,
...
+
+        is_output_created = False
+        n_grp_values = len(actual_grp_level_counts)
+        grp_cols_list = grp_col_str.split(',') if grp_cols else []
+
+        # for each group
+        for grp_vals, actual_level_counts in actual_grp_level_counts.items():
+            target_class_sizes = _get_target_level_counts(
+                sampling_strategy_str, parsed_class_sizes,
+                actual_level_counts, output_table_size, n_grp_values)
```
njayaram2: If the `sampling_strategy_str` contains a class level that does not exist in some group, the current implementation fails when this for loop starts processing that particular group.
Can we change `_get_target_level_counts` to work with `actual_grp_level_counts`, that returns back a group specific `target_class_sizes` instead (say something like `target_grp_class_sizes`)? That way, the failure would happen much earlier, before SQL sampling queries are run for other groups.

If I am not mistaken, we can create a new function that would call `_get_target_level_counts` in a loop, but returns back a dict with `grp_vals` being the key, and `target_class_sizes` being the values.

src/ports/postgres/modules/sample/balance_sample.py_in
```diff
@@ -468,81 +544,107 @@ def balance_sample(schema_madlib, source_table, output_table, class_col,
...
+
```
njayaram2: Following the above comment, we can then have a new for loop that starts here, which will be executed only if there is no error condition arising due to `sampling_strategy_str`.

src/ports/postgres/modules/sample/balance_sample.py_in
```diff
@@ -58,28 +60,64 @@ NOSAMPLE = 'nosample'
...
+            -- for each group and class combination present in source table
+            -- get the count of rows for that combination
+            SELECT
+                {array_grp_by_cols_comma}
+                ({class_col})::TEXT AS classes,
+                count(*) AS class_count
+            FROM {source_table}
+            GROUP BY {grp_by_cols_comma} ({class_col})
+        ) q
+        GROUP BY {grp_identifier}
```
njayaram2: In Greenplum 5.x, the above query fails on install check, with the following error:
```
SELECT balance_sample('"TEST_s"', 'out_sr2', 'gr1', 'undersample ', NULL, NULL, TRUE, TRUE);
psql:/tmp/madlib.2N5sjK/sample/test/balance_sample.sql_in.tmp:111: ERROR:  plpy.SPIError: non-integer constant in GROUP BY
LINE 17:         GROUP BY true
                          ^
QUERY:
        SELECT
            -- For each group get the classes and their rows counts
            true as group_values,
            array_agg(classes) as classes,
            array_agg(class_count) as class_count
        FROM(
            -- for each group and class combination present in source table
            -- get the count of rows for that combination
            SELECT

                (gr1)::TEXT AS classes,
                count(*) AS class_count
            FROM "TEST_s"
            GROUP BY  (gr1)
        ) q
        GROUP BY true

  PL/Python function "balance_sample", line 23, in <module>
    return balance_sample.balance_sample(**globals())
  PL/Python function "balance_sample", line 575, in balance_sample
  PL/Python function "balance_sample", line 100, in _get_level_frequency_distribution
PL/Python function "balance_sample"
```

iyerr3: Thanks for checking this! I've removed the constant group by clause to avoid this.

src/ports/postgres/modules/sample/balance_sample.py_in
iyerr3: I've moved the `_get_target_level_counts` outside the loop to catch the error earlier. Hopefully, all the grouping logic doesn't complicate the code much.

asfgit:
https://builds.apache.org/job/madlib-pr-build/362/

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +545,95 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+DROP TABLE IF EXISTS output_table;
+SELECT madlib.balance_sample(
+    'flags',          -- Source table
+    'output_table',   -- Output table
+    'mainhue',        -- Class column
+    'blue=8',         -- Specified class value size. Rest of the values are outputed as is.
+    NULL,             -- Output table size
+    'zone'            -- Group by zone
+);
+SELECT * FROM output_table ORDER BY zone, mainhue;
```
fmcquillan99: For some reason, doxygen builds the above as
```
DROP TABLE IF EXISTS output_table;
SELECT madlib.balance_sample(
    'flags,          -- Source table
    'output_table',   -- Output table
    'mainhue',        -- Class column
    'blue=8',         -- Specified class value size. Rest of the values are outputed as is.
    NULL,             -- Output table size
    'zone'            -- Group by zone
);
SELECT * FROM output_table ORDER BY zone, mainhue;
```
where flags is missing the trailing '

src/ports/postgres/modules/sample/balance_sample.sql_in
```diff
@@ -543,6 +545,95 @@ SELECT * FROM output_table ORDER BY mainhue, name;
...
+             3 |  5 | Canada      |        1 |    4 | 9976 |         24 |        1 |       2 | red
+             7 | 15 | Portugal    |        3 |    4 |   92 |         10 |        6 |       5 | red
+             8 | 16 | Spain       |        3 |    4 |  505 |         38 |        2 |       2 | red
+             9 | 19 | UK          |        3 |    4 |  245 |         56 |        1 |       3 | red
+            10 | 20 | USA         |        1 |    4 | 9363 |        231 |        1 |       3 | white
+             4 | 10 | Ireland     |        3 |    4 |   70 |          3 |        1 |       3 | white
+(23 rows)
+</pre>
+
 @anchor literature
```
fmcquillan99: Are we supporting specifying the number of nulls?
```
DROP TABLE IF EXISTS output_table;
SELECT madlib.balance_sample(
    'flags',          -- Source table
    'output_table',   -- Output table
    'mainhue',        -- Class column
    'null=8',         -- Specified class value size. Rest of the values are outputed as is.
    NULL,             -- Output table size
    'zone' ,           -- Group by zone
    NULL,             -- With replacement
    TRUE              -- Keep null
);
SELECT * FROM output_table ORDER BY zone, mainhue;
```
produces
```
/Users/fmcquillan/anaconda/lib/python2.7/site-packages/sqlalchemy/engine/default.pyc in do_execute(self, cursor, statement, parameters, context)
    448 
    449     def do_execute(self, cursor, statement, parameters, context=None):
--> 450         cursor.execute(statement, parameters)
    451 
    452     def do_execute_no_params(self, cursor, statement, context=None):

InternalError: (psycopg2.InternalError) plpy.Error: Invalid input param name: null (plpython.c:4960)
  PL/Python function "balance_sample", line 23, in <module>
    return balance_sample.balance_sample(**globals())
  PL/Python function "balance_sample", line 559, in balance_sample
  PL/Python function "balance_sample", line 604, in extract_keyvalue_params
PL/Python function "balance_sample"
 [SQL: "SELECT madlib.balance_sample(\n    'flags',          -- Source table\n    'output_table',   -- Output table\n    'mainhue',        -- Class column\n    'null=8',         -- Specified class value size. Rest of the values are outputed as is.\n    NULL,             -- Output table size\n    'zone' ,           -- Group by zone\n    NULL,\n    TRUE\n);"]
```

njayaram2: This is an expected failure. We explicitly check for such conditions and fail in our utility function that extracts key value params (`extract_keyvalue_params()`):
`if not param_name or param_name in ('none', 'null'):`

fmcquillan99: That's OK, thanks for checking.

asfgit:
https://builds.apache.org/job/madlib-pr-build/363/

asfgit:
https://builds.apache.org/job/madlib-pr-build/365/

asfgit:
https://builds.apache.org/job/madlib-pr-build/366/

asfgit:
https://builds.apache.org/job/madlib-pr-build/367/

asfgit:
https://builds.apache.org/job/madlib-pr-build/368/

PR #240, opened 2018-03-10
MLP: Fix step size initialization based on learning rate policy

njayaram2:
JIRA: MADLIB-1212

The step_size is supposed to be updated based on the learning rate. The
formulae for different policies depend on the current iteration number,
which was not consumed correctly. This commit fixes it by using
it.iteration for the current iteration number in that update.

Co-authored-by: Nikhil Kak <nkak@pivotal.io>

### DIFF
--- a/src/ports/postgres/modules/convex/mlp_igd.py_in
+++ b/src/ports/postgres/modules/convex/mlp_igd.py_in
@@ -112,6 +112,7 @@ def mlp(schema_madlib, source_table, output_table, independent_varname,
     num_output_nodes = 0
     classes = []
     dependent_type = get_expr_type(dependent_varname, source_table)
+    classes_str = None
     original_dependent_varname = dependent_varname
 
     x_mean_table = unique_string(desp='x_mean_table')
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/364/

src/ports/postgres/modules/convex/mlp_igd.py_in
```diff
@@ -112,6 +112,7 @@ def mlp(schema_madlib, source_table, output_table, independent_varname,
     num_output_nodes = 0
     classes = []
     dependent_type = get_expr_type(dependent_varname, source_table)
+    classes_str = None
```
kaknikhil: we don't really need this `classes_str` variable, right ?

asfgit:
https://builds.apache.org/job/madlib-pr-build/381/

jingyimei:
We manually tested the bug fix by printing the step_size and step_size_init for each iteration and then called mlp with all the possible policies.
step_size was updated as expected.  +1 for the changes

PR #241, opened 2018-03-14
MiniBatch Pre-Processor: Add new module minibatch_preprocessing

kaknikhil:
JIRA: MADLIB-1200

MiniBatch Preprocessor is a utility function to pre-process the input
data for use with models that support mini-batching as an optimization.

The main purpose of the function is to prepare the training data for minibatching algorithms.
1. If the dependent variable is boolean or text, perform one hot encoding.  N/A for numeric.
2. Typecast independent variable to double precision[]
2. Based on the buffer size, group all the dependent and independent variables in a single tuple representative of the buffer.

Notes
1. Ignore null values in independent and dependent variables
2. Standardize the input before packing it.

Other changes:
1. Removed __ from public methods in utils_regularization.py
Renamed __utils_ind_var_scales and __utils_ind_var_scales_grouping
so that we can access them from within a class, more specifically
the minibatch_preprocessing module.
2. Added new function for regex match and refactored elastic_net.py_in to use this function

Co-authored-by: Rahul Iyer <riyer@apache.org>
Co-authored-by: Jingyi Mei <jmei@pivotal.io>

### DIFF
--- a/src/ports/postgres/modules/utilities/mean_std_dev_calculator.py_in
+++ b/src/ports/postgres/modules/utilities/mean_std_dev_calculator.py_in
@@ -1,0 +1,54 @@
+# coding=utf-8
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""
+@file mean_std_dev_calculator.py_in
+
+@brief
+
+@namespace utilities
+
+"""
+
+from convex.utils_regularization import utils_ind_var_scales
+from utilities import _array_to_string
+
+m4_changequote(`<!', `!>')
+
+#TODO: use this for all the modules that calculate the std dev and mean for x
+# mlp, pca, elastic_net
+class MeanStdDevCalculator:
+    def __init__(self, schema_madlib, source_table, indep_var_array_str, dimension):
+        self.schema_madlib= schema_madlib
+        self.source_table= source_table
+        self.indep_var_array_str = indep_var_array_str
+        self.dimension = dimension
+
+    def get_mean_and_std_dev_for_ind_var(self):
+        set_zero_std_to_one = True
+
+        x_scaled_vals = utils_ind_var_scales(self.source_table,
+                                             self.indep_var_array_str,
+                                             self.dimension,
+                                             self.schema_madlib,
+                                             None, # do not dump the output to a temp table
+                                             set_zero_std_to_one)
+        x_mean_str = _array_to_string(x_scaled_vals["mean"])
+        x_std_str = _array_to_string(x_scaled_vals["std"])
+
+        return x_mean_str, x_std_str
--- a/src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
+++ b/src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
@@ -1,0 +1,581 @@
+# coding=utf-8
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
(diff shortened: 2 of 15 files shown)


src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -0,0 +1,559 @@
...
+            num_missing_rows_skipped,
+            self.output_summary_table)
+
+    def _validate_minibatch_preprocessor_params(self):
+        # Test if the independent variable can be typecasted to a double precision
+        # array and let postgres validate the expression
+
+        # Note that this will not fail for 2d arrays but the standardizer will
+        # fail because utils_normalize_data will throw an error
+        typecasted_ind_varname = "{0}::double precision[]".format(
```
kaknikhil: This code is also duplicated in get_indep_var_array_str(). We kept the duplication because it looked cleaner.
Suggestions are welcome.

asfgit:
https://builds.apache.org/job/madlib-pr-build/369/

asfgit:
https://builds.apache.org/job/madlib-pr-build/370/

asfgit:
https://builds.apache.org/job/madlib-pr-build/373/

asfgit:
https://builds.apache.org/job/madlib-pr-build/374/

asfgit:
https://builds.apache.org/job/madlib-pr-build/375/

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -0,0 +1,559 @@
...
+    def minibatch_preprocessor_help(schema_madlib, message):
+        method = "minibatch_preprocessor"
+        summary = """
+        ----------------------------------------------------------------
+                            SUMMARY
+        ----------------------------------------------------------------
+        MiniBatch Preprocessor is a utility function to pre process the input
+        data for use with models that support mini-batching as an optimization
+
+        #TODO add more here
```
njayaram2: Is the comment still required?

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -0,0 +1,559 @@
...
+                   source_table = source_table,
+                   output_table = output_table,
+                   dependent_varname = dep_var_array_str,
+                   independent_varname = indep_var_array_str,
+                   buffer_size = buffer_size,
+                   class_values = class_values,
+                   num_rows_processed = num_rows_processed,
+                   num_missing_rows_skipped = num_missing_rows_skipped,
+                   grouping_cols = "NULL")
+        plpy.execute(query)
```
njayaram2: I tried running the module, without setting any value for `buffer_size`. In the output table, each buffer had exactly one row (due to the work in progress formula used in `MiniBatchBufferSizeCalculator`), but the summary table reported `buffer_size = 38.0`.
`buffer_size` in the output table and summary table are in sync when I specify a value for `buffer_size`, but there seems to be an issue when we don't specify the `buffer_size`.

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -0,0 +1,559 @@
...
+
+        # Note that this will not fail for 2d arrays but the standardizer will
+        # fail because utils_normalize_data will throw an error
+        typecasted_ind_varname = "{0}::double precision[]".format(
+                                                    self.independent_varname)
+        validate_module_input_params(self.source_table, self.output_table,
+                                     typecasted_ind_varname,
+                                     self.dependent_varname, self.module_name)
+
+        self._validate_other_output_tables()
```
njayaram2: This is linked to the last comment on this PR. We could use an optional param in `validate_module_input_params` to validate other output tables too (a list of suffixes such as `['_summary', '_standardization']`).

njayaram2:
Another issue I found but forgot to mention in the review:
The `__id__` column has double values instead of integers. For instance, I found values such as `0.20000000000000000000` for that column in the output table.
This issue also happens only when the module is used without specifying a value for the `buffer_size` param.

src/ports/postgres/modules/utilities/mean_std_dev_calculator.py_in
```diff
@@ -0,0 +1,54 @@
...
+
+"""
+
+from convex.utils_regularization import utils_ind_var_scales
+from utilities import _array_to_string
+
+m4_changequote(`<!', `!>')
+
+#TODO: use this for all the modules that calculate the std dev and mean for x
+# mlp, pca, elastic_net
```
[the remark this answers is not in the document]
jingyimei: Yes the TODO comment is left here for future reference. The idea is that whenever we update any of these modules, we can refactor the code to call this class.

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
jingyimei: Yes. This is a placeholder for adding more detail description later.

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
jingyimei: good catch. The problem is that we assumed that math.ceil returns an int but it returns a float and hence the bug. Our unit tests could not catch this because comparing an int value for a float variable is valid. We need to assert that the instance type of the variable is int. 
We will improve the unit test as well as fix the bug.

jingyimei:
@njayaram2 
```
Another issue I found but forgot to mention in the review:
The __id__ column has double values instead of integers. For instance, I found values such as 0.20000000000000000000 for that column in the output table.
This issue also happens only when the module is used without specifying a value for the buffer_size param.
```

This is the caused by the same math.ceil bug as mentioned above. will take care of this.

asfgit:
https://builds.apache.org/job/madlib-pr-build/388/

asfgit:
https://builds.apache.org/job/madlib-pr-build/389/

asfgit:
https://builds.apache.org/job/madlib-pr-build/390/

asfgit:
https://builds.apache.org/job/madlib-pr-build/393/

asfgit:
https://builds.apache.org/job/madlib-pr-build/394/

PR #249, opened 2018-03-26
RF: Use NULL::integer[] when no continuous features

iyerr3:
JIRA: MADLIB-1219

When variable importance is enabled, to compute importance score,
distribution of the categorical and continuous features are computed.
For continuous features, this function requires initializing a vector of
length = number of continuous features. When there are no continuous
features, this initialization fails.

This commit fixes the issue by inputing a NULL::integer[] vector when
there no continuous features.

Closes #249

### DIFF
--- a/src/ports/postgres/modules/recursive_partitioning/random_forest.py_in
+++ b/src/ports/postgres/modules/recursive_partitioning/random_forest.py_in
@@ -946,6 +946,11 @@ def _calculate_oob_prediction(
 
     oob_var_dist_view = unique_string()
     if importance:
+        if con_features:
+            initialized_float_array = ("{0}.array_of_float({1})::integer[]".
+                                       format(schema_madlib, len(con_features)))
+        else:
+            initialized_float_array = "NULL::integer[]"
         sql_create_oob_var_dist_view = """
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/402/

asfgit:
https://builds.apache.org/job/madlib-pr-build/403/

fmcquillan99:
See https://issues.apache.org/jira/browse/MADLIB-1219 for results from my tests.

LGTM

asfgit:
https://builds.apache.org/job/madlib-pr-build/406/

kaknikhil:
The changes look good. +1 for adding an install check test.

I noticed that `con_features` is populated by `decision_tree.py:_classify_features` which does something like `con_types = ['real', 'float8', 'double precision', 'numeric']`. Is it better to call the utilities function `is_psql_numeric_type`  ?

iyerr3:
@kaknikhil Thanks for the suggestion. It makes sense to use that function with integer types excluded (since DT does not treat integer as continuous). I'll add that as a separate commit since it's unrelated to this PR.

PR #259, opened 2018-04-10
Minibatch: Add one-hot encoding option for int

iyerr3:
JIRA: MADLIB-1226

Integer dependent variables can be used either in regression or
classification. To use in classification, they need to be one-hot
encoded. This commit adds an option to allow users to pick if a integer
dependent input needs to one-hot encoded or not. The flag is ignored if
the variable is not of integer type.

Other changes include adding an appropriate test in install-check,
code cleanup and PEP8 conformance.

### DIFF
--- a/src/ports/postgres/modules/utilities/control.py_in
+++ b/src/ports/postgres/modules/utilities/control.py_in
@@ -10,19 +10,41 @@ m4_changequote(`<!', `!>')
 
 @brief driver functions shared by modules
 """
-
-from distutils.util import strtobool
 import plpy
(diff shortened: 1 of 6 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/431/

iyerr3:
I've (force) pushed after the rebase. This should now reflect the `dependent_vartype` change from previous PR.

asfgit:
https://builds.apache.org/job/madlib-pr-build/444/

PR #260, opened 2018-04-10
minibatch preprocessor improvements

kaknikhil:
This PR makes two improvements to the preprocessor code

1. Check for all character types for dependent col
2. Create temp table for standardization.

See the commit for more details


src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -397,8 +408,9 @@ class MiniBatchStandardizer:
             x_std_dev_str = self.x_std_dev_str)
         return query
 
-    def _get_query_for_standardizing_with_grouping(self):
+    def _create_table_for_standardizing_with_grouping(self):
```
njayaram2: Why was the method name changed? The older name seems to be more apt, since this function is still returning the query, and not executing it (the same for `_create_table_for_standardizing_without_grouping()` too).

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -387,6 +397,7 @@ class MiniBatchStandardizer:
         ) as {ind_colname}
         FROM {source_table}
         """.format(
+            standardized_table = self.standardized_table,
```
njayaram2: I think the following also works (using `self` inside format), and looks neater:
```
query="""
        CREATE TEMP TABLE {self.standardized_table} as
        SELECT
        {self.dep_var_array_str} as {dep_colname},
        {self.schema_madlib}.utils_normalize_data
        (
            {self.indep_var_array_str},'{self.x_mean_str}'::double precision[],
            '{self.x_std_dev_str}'::double precision[]
        ) as {ind_colname}
        FROM {self.source_table}
        """.format(self=self,
                   dep_colname = MINIBATCH_OUTPUT_DEPENDENT_COLNAME,
                   ind_colname = MINIBATCH_OUTPUT_INDEPENDENT_COLNAME)
```
This could be applied in several other queries too in this file.

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
kaknikhil: yes you are right. It was initially creating the table inside the function and later I moved it outside but forgot to rename the function. will update the name

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
kaknikhil: yes, #259 also made a few changes related to this. Will update all that are remaining

asfgit:
https://builds.apache.org/job/madlib-pr-build/448/

PR #261, opened 2018-04-11
MLP: Check for 1-hot encoding of dependent variable for minibatch

orhankislal:
This commit adds a check to make sure that the dependent variable for mlp
minibatch is one hot encoded. This only validates that the dependent
variable array has more than 1 value.


asfgit:
https://builds.apache.org/job/madlib-pr-build/439/

PR #263, opened 2018-04-17
Bugfix/mlp minibatch grouping

iyerr3:
If mini-batch preprocessor is run with grouping, the standardization in the output table is computed per group. This implies that MLP should also be run with the same grouping, else the dataset used for training would be different from the original data, hence making the training invalid. 

This commit ensures that MLP training will proceed only if the grouping column input is same as the one used during preprocessing.


fmcquillan99:
I tested this quite a bit and it seems to work nicely for me.

LGTM

PR #275, opened 2018-06-05
Madpack: Fix error with dropping user after IC failure.

njayaram2:
JIRA: MADLIB-1182

Previously, when install check did not fail gracefully, the user created
by madpack hung around and disturbed IC attempts within other databases.
We fixed this by:
1) Renaming the test user using the specific database that the IC test
was run from, making the test user name database-specific.
2) Dropping schema and user in a try-finally block, so that it's executed even
on a failed IC run.

Co-authored-by: Arvind Sridhar <asridhar@pivotal.io>

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -822,9 +822,10 @@ def run_install_check(args, testcase):
         return
 
     # Create install-check user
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/497/

PR #276, opened 2018-06-07
Feature/dev check

orhankislal:


asfgit:
https://builds.apache.org/job/madlib-pr-build/499/

src/madpack/madpack.py
```diff
@@ -898,13 +900,19 @@ def run_install_check(args, testcase):
...
+            sql_files = maddir_mod_sql + '/' + module + '/test/*.ic.sql_in'
+        else:
+            sql_files = maddir_mod_sql + '/' + module + '/test/*.sql_in'
         for sqlfile in sorted(glob.glob(sql_files), reverse=True):
             algoname = os.path.basename(sqlfile).split('.')[0]
             # run only algo specified
             if (module in modset and modset[module] and
                     algoname not in modset[module]):
                 continue
+            # Do not run test/*.ic.sql_in files for dev-check.
```
iyerr3: I believe if you use `*[!ic].sql_in` in the dev-check glob expression, it will exclude the ic files. You won't need these lines if that works.

njayaram2: That's great, will try it out.

njayaram2:
Thank you for the comments @iyerr3 , will make the changes you have requested.

Having one IC file for each module makes sense, but on Greenplum, for some modules the IC run time is still quite high. For example, if a module had 5 IC files, where each ran for 10 seconds, the user would see IC progressing every 10 seconds. But if those were combined into one IC file, the progression would happen after 50 seconds. It seemed a little odd (longer IC run times for some modules in terms of user experience) when we were trying it out, so decided to keep multiple IC files for such modules.

asfgit:
https://builds.apache.org/job/madlib-pr-build/517/

asfgit:
https://builds.apache.org/job/madlib-pr-build/518/

asfgit:
https://builds.apache.org/job/madlib-pr-build/520/

PR #278, opened 2018-06-11
Upgrade: add changelist file and bug fixes

kaknikhil:
This PR  adds a changelist file for testing 1.14 to 1.15 upgrade and addresses 2 bugs we found during upgrade testing.
1. Commit 8e34f68 added a new function called `_write_to_file` that takes 2 arguments.
Some of the calls to this function were not passing the first file handle argument.
2. Append schema_madlib to the mlp_igd_final return type. This caused the
upgrade to fail from 1.12 to 1.x if there was a dependency on
mlp_igd_final.

Co-authored-by : Orhan Kislal <okislal@pivotal.io>

### DIFF
--- a/src/madpack/changelist_1.12_1.13.yaml
+++ b/src/madpack/changelist_1.12_1.13.yaml
@@ -49,7 +49,7 @@ udf:
         rettype: void
         argument: character varying, character varying, character varying, character varying
     - mlp_igd_final:
-        rettype: mlp_step_result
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/507/

kaknikhil:
@iyerr3 yes we will have to rename it.

PR #281, opened 2018-06-21
Support special characters in MLP, minibatch preprocessor and encode_categorical

kaknikhil:
Support special characters in MLP, minibatch preprocessor and encode_categorical

JIRA: MADLIB-1237
JIRA: MADLIB-1238
JIRA: MADLIB-1238
JIRA: MADLIB-1243

The module that needs to support special characters will have to call
quote_literal() on all the column values that need to be escaped and
quoted and then this list can be passed to the py_list_to_sql_string
function

We also created a function called get_distinct_col_levels which will
call quote_literal and then return a list of escaped column levels. The
output of this function can then be safely passed to
py_list_to_sql_string with long_format set as True.

Co-Authored-by: Jingyi Mei <jmei@pivotal.io>
Co-Authored-by: Rahul Iyer <riyer@apache.org>
Co-Authored-by: Arvind Sridhar <asridhar@pivotal.io>


asfgit:
https://builds.apache.org/job/madlib-pr-build/521/

asfgit:
https://builds.apache.org/job/madlib-pr-build/522/

PR #284, opened 2018-06-27
SVM: Fix flaky dev-check failure

njayaram2:
JIRA: MADLIB-1232

SVM has a dev-check query that is flaky on a large cluster. This commit
relaxes the assert condition for that query.

Closes #284

### DIFF
--- a/src/ports/postgres/modules/svm/test/svm.sql_in
+++ b/src/ports/postgres/modules/svm/test/svm.sql_in
@@ -214,7 +214,10 @@ SELECT * FROM svm_model_small_norm2;
 SELECT
     assert(
         norm2(l2.coef) < norm2(noreg.coef) OR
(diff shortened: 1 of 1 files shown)


iyerr3:
This test isn't necessarily the best thing to check but if we do want to keep it with some relaxed constraints then maybe we allow the norm to go higher by a small amount iff the total loss is lower. So something along the lines of 
```
norm2(l2.coef) < norm2(noreg.coef) OR
    ( (norm2(l2.coef)-norm2(noreg.coef))/norm2(noreg.coef) < 0.1 AND l2.loss < noreg.loss),
```

njayaram2:
Thank you for the comment @iyerr3 , will relax the constraint as suggested.

asfgit:
https://builds.apache.org/job/madlib-pr-build/528/

PR #286, opened 2018-07-03
Build: Remove symlinks during rpm uninstall

iyerr3:
JIRA: MADLIB-1175

`rpm --install` creates three symlinks to `Versions/`, `.../bin`, and
`.../doc`. These symlinks should be deleted during `rpm --erase`.
Additionally, we also delete `Versions/` if it is empty after the erase.

Closes #286

Co-Authored-by: Arvind Sridhar <asridhar@pivotal.io>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
               <exclude>deploy/preflight.sh</exclude>
               <exclude>deploy/RPM/CMakeLists.txt</exclude>
               <exclude>deploy/rpm_post.sh</exclude>
+              <exclude>deploy/rpm_post_uninstall.sh</exclude>
               <exclude>doc/bin/CMakeLists.txt</exclude>
               <exclude>doc/bin/doxypy.py</exclude>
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/532/

pom.xml
```diff
@@ -66,6 +66,7 @@
               <exclude>deploy/preflight.sh</exclude>
               <exclude>deploy/RPM/CMakeLists.txt</exclude>
               <exclude>deploy/rpm_post.sh</exclude>
+              <exclude>deploy/rpm_post_uninstall.sh</exclude>
```
orhankislal: Since this is a new file, we should add the apache license instead of adding it to the exclude list.

iyerr3: This was done under the assumption that Apache does not need the license header for simple scripts that don't have considerable IP. In this case, we're only removing some files and a directory, so assuming that it's fine to not explicitly tag the license. Let me know if you feel otherwise.

PR #291, opened 2018-07-11
Feature: Vector-Column Transformations

ArvindSridhar:
JIRA: MADLIB-1240

This commit adds a new SQL function called vec2cols and refactors the
current function cols2vec, providing greater integration between the two
modules. We now have a single Python file with separate classes for each
feature. We also have unified unit-tests and dev-check/install-check
tests.

The vec2cols function enables users to split up a single column into
multiple columns, given that the input column contains array entries.
For example, if the input column contained ARRAY[1, 2, 3] in one of its
rows, the output table will contain 3 different columns, one for each
element of the array.

PR #292, opened 2018-07-12
Internal: Add function to check column type for 1D array

ArvindSridhar:
Added a db_utils function to check whether a column of type array contains 1D arrays only

### DIFF
--- a/src/ports/postgres/modules/internal/db_utils.py_in
+++ b/src/ports/postgres/modules/internal/db_utils.py_in
@@ -79,3 +79,12 @@ def quote_literal(input_str):
         return "{qd}{input_str}{qd}".format(qd=QUOTE_DELIMITER,
                                             input_str=input_str)
 # ------------------------------------------------------------------------------
+
+def is_col_1d_array(source_table, col_name):
+    query = """
+        SELECT array_upper({0}, 2) IS NULL AS n_y
+        FROM {1}
+        LIMIT 1
+    """.format(col_name, source_table)
+    result = plpy.execute(query)
+    return result[0]["n_y"]


asfgit:
https://builds.apache.org/job/madlib-pr-build/553/

asfgit:
https://builds.apache.org/job/madlib-pr-build/554/

iyerr3:
Few comments: 
1. The query would read better if it is setup to return a boolean i.e. using `array_upper(..., 2) IS NULL` instead of converting the integer output to bool in Python. 
2. My request in #291 for moving this to a separate commit was for the `plpy_mock.py_in` file since that looks like an independent work. This function is minor enough that it can be kept in #291. 
3. The unit test for this function isn't really testing anything except the last line in the function. That also becomes moot if comment 1. is accepted. Considering this, I suggest we don't have the unit test for this function (at least not in the form it currently is).

ArvindSridhar:
I changed the query but was wondering: since we are now returning just result[0]['n_y'] in python, could this cause problems if our functions calling this expect a python boolean output, while this function outputs a SQL boolean which gets converted into a python string?

iyerr3:
That's a valid concern. You could return bool(result[0]['n_y']) to ensure a
bool output instead of a string (though I believe the SPI ensures that
boolean from SQL is returned as bool in Python).

>

iyerr3:
Yeah, the unit tests are unnecessary after this change. I vote to remove
them, with just the function added back to the original PR.

ArvindSridhar:
Sounds good. Just pushed a commit to this PR, will rebase and add back the original function to the original vec2cols PR tomorrow

asfgit:
https://builds.apache.org/job/madlib-pr-build/557/

asfgit:
https://builds.apache.org/job/madlib-pr-build/558/

asfgit:
https://builds.apache.org/job/madlib-pr-build/559/

ArvindSridhar:
PR ready to be closed, changes made elsewhere already

PR #295, opened 2018-07-18
Recursive Partitioning: Add function to report importance scores

njayaram2:
JIRA: MADLIB-925

This commit adds a new MADlib function (get_var_importance) to report the
importance scores in decision tree and random forest. RF models prior to
MADlib 1.15 used to have variable importance scores reported, but they
also have impurity variable importance from 1.15 onwards. This function
reports both those scores for >=1.15 RF models, and only the oob variable
importance score for <1.15 RF models.
This function when called for a DT model, would return the impurity
variable importance score for >=1.15 DT models.

Co-authored-by: Jingyi Mei <jmei@pivotal.io>


asfgit:
https://builds.apache.org/job/madlib-pr-build/573/

asfgit:
https://builds.apache.org/job/madlib-pr-build/576/

src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
```diff
@@ -2327,6 +2328,110 @@ def _tree_error(schema_madlib, source_table, dependent_varname,
...
+def _is_RF_model_with_imp_pre_1_15(group_table, summary_table):
+    """
+        Check if the RF model is from MADlib < 1.15. The group table for
+        >= 1.15 RF models should have a column named impurity_var_importance
+        if it was learnt with importance param True.
+    """
+    _assert(table_exists(group_table),
+        "Recursive Partitioning: Model group table does not exist.")
+    # this flag has to be set to true for RF to report importance scores.
+    isImportance = plpy.execute("SELECT importance FROM {summary_table}".
```
iyerr3: Our convention is to use snake case for Python (i.e. `is_importance`). I also suggest changing it to `is_importance_set` to make it more explicit.

src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
```diff
@@ -2327,6 +2328,110 @@ def _tree_error(schema_madlib, source_table, dependent_varname,
...
+    _assert(table_exists(model_table),
+            "Recursive Partitioning: Model table does not exist.")
+    _assert(table_exists(summary_table),
+            "Recursive Partitioning: Model summary table does not exist.")
+    _assert(not table_exists(output_table),
+            "Recursive Partitioning: Output table already exists.")
+
+def _is_model_for_RF(summary_table):
+    # Only an RF model (and not DT) would have num_trees column in summary
+    return columns_exist_in_table(summary_table, ['num_trees'])
```
iyerr3: We've a `method_name` in summary table that makes this clearer.

src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
```diff
@@ -2327,6 +2328,110 @@ def _tree_error(schema_madlib, source_table, dependent_varname,
...
+
+def _is_RF_model_with_imp_pre_1_15(group_table, summary_table):
```
iyerr3: Since goal of function is to check if `impurity_var_importance` exists in group table, let's name it to reflect that.

src/ports/postgres/modules/recursive_partitioning/decision_tree.py_in
```diff
@@ -2327,6 +2328,110 @@ def _tree_error(schema_madlib, source_table, dependent_varname,
...
+        if is_pre_1_15_RF_model:
+            # In < 1.15 RF model, the variable importance was captured using two
+            # different columns named 'cat_var_importance'
+            plpy.execute(
+                """ CREATE TABLE {output_table} AS
+                    SELECT {grouping_cols_comma}
+                        unnest(regexp_split_to_array(cat_features, ',')) AS feature,
+                        unnest(cat_var_importance) AS var_importance
+                    FROM {group_table}, {summary_table}
+                """.format(**locals()))
```
iyerr3: IMO, it's cleaner to see the two queries `UNION`ed together to get the output.

njayaram2:
Thank you for the comments @iyerr3 , will make necessary changes.

asfgit:
https://builds.apache.org/job/madlib-pr-build/580/

fmcquillan99:
Should impurity_var_importance always add up to 100?
From the regression example in the user docs:

```
DROP TABLE IF EXISTS mt_imp_output;
SELECT madlib.get_var_importance('mt_cars_output','mt_imp_output');
SELECT am, impurity_var_importance FROM mt_imp_output ORDER BY am, impurity_var_importance DESC;
```
results in
```

 am | impurity_var_importance 
----+-------------------------
  0 |        35.7664683110879
  0 |        24.7481977075922
  0 |        12.4401197123678
  0 |        12.1559096708347
  0 |        4.88929809351791
  1 |        31.7259035495099
  1 |        29.6146492693988
  1 |        14.9602257795489
  1 |        7.01369118455985
  1 |        6.68552870777581
(10 rows)
```
which does not add up to 100
```
		grp 0			grp 1
		35.76646831		31.72590355
		24.74819771		29.61464927
		12.44011971		14.96022578
		12.15590967		7.013691185
		4.889298094		6.685528708
total	89.9999935		89.99999849
```

fmcquillan99:
Another run I got
```
			grp 0				grp1
			31.01364943			31.66666576
			22.85881741			33.33333245
			13.70257438			0
			6.344527751			3.333333304
			26.0804244			11.66666654
total		99.99999336			79.99999806
```
so this does seem to be about trees contributing or not.

njayaram2:
@fmcquillan99 the importance scores for all variables within a tree would sum up to 100. But in the case of a forest, we average over the scores for each variable across all trees in the forest, and the sum of those averages might not be 100. One reason could be because some variables might get a importance score of `0` in some trees, and with trees that have only a single node, all the variables would get a score of `0`.

iyerr3:
Considering the above situation, I suggest the variable importance values not be scaled to sum to 100. We can make the normalization within `get_var_importance` just for the reporting (which is the behavior in rpart). In other words, the output table would keep the original values (for DT and RF) but the helper function would rescale during the report for ease in reading the values.

fmcquillan:
Would this apply to oob too?
Or just impurity?

njayaram2:
@iyerr3, as you noted, that would change the importance scores that appear in the DT/RF output table, and the `get_var_importance` output table. Wouldn't that be a little confusing?
One other option is to do the scaling outside the C++ function, inside python just before writing it to the DT/RF output table (although this would force us to call that normalization code in both DT and RF).

njayaram2:
@fmcquillan only impurity, I don't think we scale oob to 100.

iyerr3:
So the way R does this is it keeps the original values in its model. But it has a "report" function that outputs multiple things including importance and in this report the values are normalized. The user can access the original values from the model if needed but would mostly use the report to understand the importance. 

We could essentially do the same and be clear about this in the documentation. Alternatively, we chuck the normalization to 100. It was done solely to compare with R.

fmcquillan99:
I like this last suggestion from @iyerr3, that we report raw values for oob and impurity VI in the model output file.  (OK to keep the shifted oob > 0 as we do now.)

For the helper/reporting function, compute and report out the scaled/normalized values 0-100 for both oob and impurity VI.  These should always add up to 100 unless there is some corner cases, if so pls let us know.

asfgit:
https://builds.apache.org/job/madlib-pr-build/591/

src/ports/postgres/modules/recursive_partitioning/random_forest.py_in
```diff
@@ -1564,69 +1578,69 @@ def get_var_importance(schema_madlib, model_table, output_table, **kwargs):
...
+
+    plpy.execute("""
+        CREATE TABLE {output_table} AS
+            SELECT {grouping_cols_comma}
+                unnest(regexp_split_to_array(independent_varnames, ',')) AS feature,
+                {oob_var_importance_str}
+                {impurity_var_importance_str}
+            FROM {importance_model_table}, {summary_table}
+        """.format(**locals()))
+# ------------------------------------------------------------------------------
```
njayaram2: +1

asfgit:
https://builds.apache.org/job/madlib-pr-build/592/

asfgit:
https://builds.apache.org/job/madlib-pr-build/601/

src/ports/postgres/modules/recursive_partitioning/random_forest.sql_in
```diff
@@ -1557,6 +1618,32 @@ m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `CONTAINS SQL', `');
...
+    output_table   TEXT
+) RETURNS VOID AS $$
+  PythonFunction(recursive_partitioning, random_forest, get_var_importance)
+$$ LANGUAGE plpythonu VOLATILE
+m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `MODIFIES SQL DATA', `');
+
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.get_var_importance(
+    message     TEXT
+) RETURNS TEXT AS $$
+  PythonFunction(recursive_partitioning, random_forest, tree_importance_help_message)
```
njayaram2: `tree_importance_help_message` is defined in `decision_tree.py_in`, so `select madlib.get_var_importance()` fails at the moment.

iyerr3: Good catch

fmcquillan99:
LGTM, here is an RF example:
```
SELECT * FROM mt_imp_output ORDER BY am, oob_var_importance DESC;
 am | feature | oob_var_importance | impurity_var_importance 
----+---------+--------------------+-------------------------
  0 | cyl     |   31.6266798136018 |        8.99888201496216
  0 | disp    |   21.3534749649495 |        30.5938284017064
  0 | vs      |   20.2312669968611 |        25.4855561460076
  0 | wt      |   16.3410741245189 |        19.7783684870616
  0 | qsec    |   10.4475041000687 |        15.1433649502623
  1 | wt      |    34.239597267579 |        24.9348163610914
  1 | disp    |   29.4316514472623 |        31.1638455198447
  1 | cyl     |   21.9435741528927 |        20.1221371309527
  1 | vs      |   14.3851771322661 |        17.5142973837102
  1 | qsec    |                  0 |        6.26490360440106
(10 rows)
```

asfgit:
https://builds.apache.org/job/madlib-pr-build/630/

PR #299, opened 2018-07-25
RF: Port DT fix for incorrect importance vector length

iyerr3:
JIRA: MADLIB-1254

Commit 0f7834e contained a fix in DT that ensured that impurity variable
importance was of the correct length for each group if a single group
eliminated a categorical variable.

This commit applies the same fix for random forest.

Closes #299


asfgit:
https://builds.apache.org/job/madlib-pr-build/598/

PR #307, opened 2018-08-06
Two CV-related changes

iyerr3:
This PR includes two commits. 

First is a fix related to issue MADLIB-1014 that led to a "cache lookup" error when the install-check schema is deleted. 

Second is refactoring of internal/cross_validation. CV had an inconsistent API that led to failure with either elastic_net or SVM. This commit simplifies the API and makes it consistent.


asfgit:
https://builds.apache.org/job/madlib-pr-build/633/

PR #310, opened 2018-08-11
Build: Download compatible Boost if version >= 1.65

iyerr3:
JIRA: MADLIB-1235

BOOST 1.65.0 removed the TR1 library which is required by MADlib till
C++11 is completely supported. Hence, we force download of a compatible
version if existing Boost is 1.65 or greater. This should be removed
when TR1 dependency is removed.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -103,21 +103,29 @@ set(MAD_MODULE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/modules)
 # -- Third-party dependencies: Find or download Boost --------------------------
 
 find_package(Boost 1.47)
-
-# We use BOOST_ASSERT_MSG, which only exists in Boost 1.47 and later.
-# Unfortunately, the FindBoost module seems to be broken with respect to version
-# checking, so we will set Boost_FOUND to FALSE if the version is too old.
 if(Boost_FOUND)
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/645/

src/CMakeLists.txt
```diff
@@ -103,21 +103,29 @@ set(MAD_MODULE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/modules)
...
     if(Boost_VERSION LESS 104600)
+        message(STATUS "No sufficiently recent version (>= 1.47) of Boost was found. Will download.")
+        set(Boost_FOUND FALSE)
+    endif(Boost_VERSION LESS 104600)
+
+    # BOOST 1.65.0 removed the TR1 library which is required by MADlib till
+    # C++11 is completely supported. Hence, we force download of a compatible
+    # version if existing Boost is 1.65 or greater. FIXME: This should be
+    # removed when TR1 dependency is removed.
+    if(NOT Boost_VERSION LESS 106500)
```
kaknikhil: maybe add to the log message that we found a boost version > 1.65 which is incompatible.

kaknikhil:
Tested with boost 1.67 and boost 1.55, works as expected. Apart from the log message suggestion, +1 from me.

PR #314, opened 2018-08-23
Ubuntu support: Enable creation of gppkg on Ubuntu

njayaram2:
This commit makes necessary changes to create a gppkg on Ubuntu. The
default behavior when MADlib is built on Ubuntu is to create a .deb
installer. If we want to create a gppkg, then we need an RPM due to
limitations in gppkg. We now create an RPM on Ubuntu (assuming package
alien is installed on Ubuntu) if the right cmake flag is specified. Once
an RPM is created on `make package`, we can now go ahead and create the
gppkg using `make gppkg`.
The cmake flag to use if we want to create an .rpm instead of .deb on
Ubuntu when we run `make package` is:
-DCREATE_RPM_FOR_UBUNTU=True

### DIFF
--- a/deploy/CMakeLists.txt
+++ b/deploy/CMakeLists.txt
@@ -26,10 +26,14 @@ elseif(UNIX)
         )
     elseif(IS_DEBIAN)
         message(STATUS "Detected Debian version ${DEB_VERSION}")
-        # -- If cmake flag -DCREATE_RPM_FOR_UBUNTU is set to some
-        # value, then we will create an RPM on Ubuntu. If that
-        # flag is not set to any value in cmake, then we create
-        # only .deb artifact on Ubuntu.
+        # By default, the cmake flag -DCREATE_RPM_FOR_UBUNTU is not
+        # set to any value. The default behavior is to create a .deb
+        # artifact when MADlib is compiled on Ubuntu.
+        # If we instead want to create an RPM artifact for MADlib
+        # on Ubuntu, then the cmake flag -DCREATE_RPM_FOR_UBUNTU can
+        # be set to some value (say True), and we will only create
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/657/

deploy/CMakeLists.txt
```diff
@@ -27,9 +26,21 @@ elseif(UNIX)
...
     elseif(IS_DEBIAN)
         message(STATUS "Detected Debian version ${DEB_VERSION}")
-        list(APPEND CPACK_GENERATOR
-            DEB
-        )
+        # -- If cmake flag -DCREATE_RPM_FOR_UBUNTU is set to some
+        # value, then we will create an RPM on Ubuntu. If that
+        # flag is not set to any value in cmake, then we create
+        # only .deb artifact on Ubuntu.
+        # Note that package alien must already be installed for
```
reductionista: The comment makes it sound as if setting -DCREATE_RPM_FOR_UBUNTU would build both .rpm and .deb while not setting it "only" builds .deb.

But looking at the code itself, only one or the other is built but never both. Assuming this is intentional, I would suggest removing "only" from the comment for clarity.

njayaram2:
Thank you for the comments @reductionista , I have updated the comment. Please do have a look at it.

asfgit:
https://builds.apache.org/job/madlib-pr-build/658/

fmcquillan99:
So this requires Alien, but we do not automatically download or bundle Alien, correct?

fmcquillan99:
Thanks @njayaram2 for the clarification.

reductionista:
Also:  a user who downloads the madlib gppkg for Ubuntu must install Alien first, before running gppkg -i on it.

PR #318, opened 2018-09-04
Madpack: Add a script for automating changelist creation

orhankislal:


asfgit:
https://builds.apache.org/job/madlib-pr-build/668/

kaknikhil:
I ran the utility with 1.13 and 1.14 as the release tags and noticed that the output file created by this utility isn't the same as the one checked in `changelist_1.13_1.14.yaml`. For ex, all the summary functions are missing from the udf section. Also the udt section has an extra entry `mlp_step_result[]:`

src/madpack/create_changelist.py
```diff
@@ -0,0 +1,132 @@
...
+# specific language governing permissions and limitations
+# under the License.
+# ------------------------------------------------------------------------------
+
+# Create changelist for any two branches
+
+# Prequisites:
+# The old version has to be installed in the "madlib_old_vers" schema
+# The new version has to be installed in the "madlib" (default) schema
+# Two branches must exist locally (run 'git fetch' to ensure you have the latest version)
```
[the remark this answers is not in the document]
orhankislal: They can be branches as well as tags.

orhankislal:
Thanks for the comments @kaknikhil. I addressed them and added the support for return type based dependencies. It would be great if you could take another look at the latest version.

asfgit:
https://builds.apache.org/job/madlib-pr-build/673/

src/madpack/diff_udf.sql
```diff
@@ -11,71 +11,13 @@ RETURNS text AS $$
...
     import plpy
     plpy.execute("""
-        CREATE TABLE functions_madlib_new_version AS
-        SELECT
-            "schema", "name", filter_schema("retype", 'madlib') retype,
-            filter_schema("argtypes", 'madlib') argtypes, "type"
-        FROM
-        (
-
-            SELECT n.nspname as "schema",
```
kaknikhil: Why was this piece of code removed ?

orhankislal: It was duplicated code. I just added another parameter to the function and reused it.

src/madpack/create_changelist.py
```diff
@@ -0,0 +1,229 @@
...
+if err1 != 0:
+    print "Schema {0} does not exist".format(old_vers)
+    raise SystemExit
+
+err1 = os.system("""psql {0} -c "select madlib.version()" > /dev/null
+                 """.format(database))
+if err1 != 0:
+    print "Schema {0} does not exist".format(new_vers)
+    raise SystemExit
+
```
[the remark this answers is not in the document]
orhankislal: That would be tricky for branches. `madlib.version()` gives a <tag-commit count-commit tag> on a branch (and just a tag on tags). We can use `git rev-parse --short HEAD` to get the commit tag but it seems complicating the code for handholding the developer that uses it.

orhankislal:
Thanks for the review @kaknikhil .

1. We had an issue with the knn help messages during one of the releases because they didn't show up on the `diff_udf.sql` output. IIRC, the problem was that the same function was converted from a pure SQL function to a plpython function. I don't have a quick solution to identify such cases.
2. This indentation should be OK.
3. We can check if the `new_vers` tag/branch exists and use `master` in its place. Alternatively, we can change the comment to avoid using the new version. It does not have any functional value.

asfgit:
https://builds.apache.org/job/madlib-pr-build/679/

orhankislal:
@kaknikhil I checked the 1.11 -> 1.12 scenario. The tool is missing the`tree_train` and `forest_train` entries. The change seems to be the removal of `surrogate_params` and the addition of `null_handling_params`. Since both of them are `text` type, it does not get picked up by the `diff_udf.sql` script.

asfgit:
https://builds.apache.org/job/madlib-pr-build/681/

src/madpack/create_changelist.py
```diff
@@ -0,0 +1,239 @@
...
+    # Find the changed types and keep a list for future use
+    os.system("psql {0} -f diff_udt.sql > /tmp/madlib_tmp_udt.txt".format(database))
+
+    f.write("\n# Changes in the types (UDT) including removal and modification\n")
+    f.write("udt:\n")
+    udt_list=[]
+    with open('/tmp/madlib_tmp_udt.txt') as fp:
+        for line in fp:
+           if 'UDT' in line and '[]' not in line:
```
kaknikhil: Can we add a comment explaining the condition in this if check ?

src/madpack/create_changelist.py
```diff
@@ -0,0 +1,239 @@
...
+    # There are two main sources for these lists.
+    # 1. The functions that actually got changed
+    # 2. The functions that depend on a changed type
+
+    # We will keep two lists (for UDF and UDA) and fill them as we parse the
+    # output of diff functions
+
+    udf_list=[]
+    uda_list=[]
+    current_list = udf_list
```
kaknikhil: Do we really need the current_list variable? The code will look a bit simpler if we just append to the right list whenever necessary.

kaknikhil:
@orhankislal The new changes look good. Apart from the 2 minor comments LGTM +1

src/madpack/create_changelist.py
orhankislal: We don't "need" the current list but it helps us to reduce the number of copy/pasted lines.

PR #319, opened 2018-09-13
Allocator: Remove 16-byte alignment for pointers in GP6

iyerr3:
Findings:
1. MADlib performs a 16-byte alignment for pointers returned by palloc.
2. Postgres prepends a small (16 byte usually) header before every
pointer which includes
	a. the memory context and
	b. the size of the memory allocation.
3. Greenplum 6+ tweaks that scheme a little: instead of the memory context,
the header tracks a "shared header" which points to another struct with
richer information (aside from the memory context).
4. Postgres calls MemoryContextContains both with the final func
for an aggregate and finalize function for a windowed aggregate.
5. Currently Postgres always concludes that the datum from MADlib is
allocated outside of the context, and makes an extra copy. In
Greenplum, MemoryContextContains needs to dereference the shared header.
This is a problem since the pointer has been shifted and the function is
misinterpreting the header.

In this commit, we disable the pointer alignment for GPDB 6+ to avoid
failure in this check. Further, we also have to disable vectorization in
Eigen since it does not work when pointers are not 16-byte aligned.

Co-authored-by: Jesse Zhang <sbjesse@gmail.com>

### DIFF
--- a/src/ports/greenplum/dbconnector/dbconnector.hpp
+++ b/src/ports/greenplum/dbconnector/dbconnector.hpp
@@ -32,7 +32,7 @@ extern "C" {
 
 #include "Compatibility.hpp"
 
-#if PG_VERSION_NUM >= 90000
+#if GP_VERSION_NUM >= 60000
     // MADlib aligns the pointers returned by palloc() to 16-byte boundaries
     // (see Allocator_impl.hpp). This is done to allow Eigen vectorization  (see
     // http://eigen.tuxfamily.org/index.php?title=FAQ#Vectorization for more
@@ -40,9 +40,8 @@ extern "C" {
     // not 16-byte aligned.
     // Further, the pointer realignment invalidates a header that palloc creates
     // just prior to the pointer address.  Greenplum after commit <> fails due
-    // to this invalid header.  Hence, the pointer realignment (and Eigen
-    // vectorization) is disabled below for Greenplum 6 and above.
-    #define DISABLE_POINTER_ALIGNMENT_FOR_GREENPLUM
+    // to this invalid header.  Hence, the pointer realignment and Eigen
+    // vectorization is disabled below for Greenplum 6 and above.
 
     // See http://eigen.tuxfamily.org/dox/group__TopicUnalignedArrayAssert.html
     // for steps to disable vectorization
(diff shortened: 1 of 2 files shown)


fmcquillan99:
So this PR only affects GP 6+ ?

It means that GP 4.3.x, GP 5 and all supported PG versions will continue to work as is, and use Eigen vectorization, if the underlying infra supports it?

iyerr3:
@fmcquillan99 Yes

> So this PR only affects GP 6+ ?
> 
> It means that GP 4.3.x, GP 5 and all supported PG versions will continue to work as is, and use Eigen vectorization, if the underlying infra supports it?

fmcquillan99:
@rahiyer doesn't this line mean vectorization is turned off for PG 9.x+ too?
`#if PG_VERSION_NUM >= 90000`

iyerr3:
The check is only in the Greenplum header
(please use @iyerr3 to mention me).

fmcquillan99:
ok, thx

LGTM

src/ports/greenplum/dbconnector/dbconnector.hpp
```diff
@@ -32,6 +32,24 @@ extern "C" {
...
+    // MADlib aligns the pointers returned by palloc() to 16-byte boundaries
+    // (see Allocator_impl.hpp). This is done to allow Eigen vectorization  (see
+    // http://eigen.tuxfamily.org/index.php?title=FAQ#Vectorization for more
+    // info).  This vectorization has to be explicitly disabled if pointers are
+    // not 16-byte aligned.
+    // Further, the pointer realignment invalidates a header that palloc creates
+    // just prior to the pointer address.  Greenplum after commit <> fails due
+    // to this invalid header.  Hence, the pointer realignment (and Eigen
+    // vectorization) is disabled below for Greenplum 6 and above.
+    #define DISABLE_POINTER_ALIGNMENT_FOR_GREENPLUM
```
kaknikhil: Do we really need this macro ? We can use the `EIGEN_DONT_VECTORIZE` macro instead.

iyerr3: Good point, I'll make the change.

asfgit:
https://builds.apache.org/job/madlib-pr-build/680/

PR #320, opened 2018-09-13
Fix false positive for Postgres 10+

d:
We used to mistake 9.3.24 as a higher version than Postgres 10 and stop
matching it to the correct "port". This patch fixes that.

### DIFF
--- a/src/ports/postgres/cmake/PostgreSQLUtils.cmake
+++ b/src/ports/postgres/cmake/PostgreSQLUtils.cmake
@@ -82,8 +82,7 @@ function(determine_target_versions OUT_VERSIONS)
                     set(VERSION "4.3ORCA")
                 endif()
             elseif(${PORT_UC} STREQUAL "POSTGRESQL" AND
-                    (${${PORT_UC}_VERSION_MAJOR} EQUAL 10 OR
-                    ${${PORT_UC}_VERSION_PATCH} GREATER 10))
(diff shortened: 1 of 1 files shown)


src/ports/postgres/cmake/PostgreSQLUtils.cmake
```diff
@@ -82,8 +82,7 @@ function(determine_target_versions OUT_VERSIONS)
                     set(VERSION "4.3ORCA")
                 endif()
             elseif(${PORT_UC} STREQUAL "POSTGRESQL" AND
-                    (${${PORT_UC}_VERSION_MAJOR} EQUAL 10 OR
-                    ${${PORT_UC}_VERSION_PATCH} GREATER 10))
```
iyerr3: The intention was to check for `...VERSION_MAJOR >= 10`. Hence the two checks - there was, however a typo with the second version input as `...PATCH` instead of `...MAJOR`. The correction should be to replace the `_PATCH` with `_MAJOR`.

iyerr3: @d Do you mind if I merge this with the minor change (as described above)?

d: oops now I understand why the `GREATER` check was there! Yeah go ahead!

PR #321, opened 2018-09-21
RF: Increase the dataset size of dev-check test

orhankislal:


fmcquillan99:
Does this fix the sporadic IC/DC issues that we have been seeing with RF?

orhankislal:
I hope so. I tried over 3000 runs with this fix in and did not get a single error.

fmcquillan99:
in that case LGTM

PR #325, opened 2018-09-27
Madpack/ic func schema

orhankislal:
IC/DC was prone to failure if the user were creating tables in the
madlib schema. This commit fixes the potential issue by removing the
madlib from the search path and adding the madlib_schema keyword for
every function, type and operator that is created by madlib.


asfgit:
https://builds.apache.org/job/madlib-pr-build/690/

jingyimei:
> @orhankislal Since this is a really big change, could you please add more info in the commit on why this change is necessary and the best available solution?

+ 1

orhankislal:
@iyerr3 @jingyimei I'll put the following in the commit message before merging.
This commit fixes the following potential issue.  
1. User deploys MADlib on the schema `madlib1`.
2. User creates a table named `vertex` in the `madlib1` schema.
3. User runs install-check.
4. The install check creates a new role and a new schema for each module in the database.
5. The install check sets the `search_path` to `madlib_installcheck_<module>, madlib1`.
6. The graph IC calls `DROP TABLE IF EXISTS vertex` and fails because the vertex table does exist but it is not owned by the install-check role.

This commit removes the madlib installation schema from the search path so that it only uses its own schema. This means every madlib function call, type and operator has to be called directly using the madlib schema name. 

One alternative solution is eliminating the `drop table` commands from the tests but that would require a very complicated refactoring work since most of the tests are written to reuse the same output table names. Another alternative is changing the `drop table` and `create table` commands to use the newly created test schema. However, this is very tricky to test; if a developer forgets to put the schema name, the test will still work unless she also creates a table of the same name in the madlib deployment schema.

iyerr3:
Wouldn't we only need to remove the first drop table call? The subsequent
ones should pick the correct one, isn't it?

I like the convenience of not having to provide the schema name in the test
and am hoping for a cleaner solution.

> One alternative solution is eliminating the drop table commands from the
> tests but that would require a very complicated refactoring work since most
> of the tests are written to reuse the same output table.
>

orhankislal:
I agree this is not a great solution. Casting the operators makes it especially awkward to use. However, we have to consider the following case. If a module has multiple test files like `graph` and if they re-use the same table names like `vertex`, then we have to drop them before re-creating.

iyerr3:
> I agree, this is not a great solution. Casting the operators makes it especially awkward to use. However, we have to consider the following case. If a module has multiple test files like `graph` and if they re-use the same table names like `vertex`, then we have to drop them before re-creating.

You mean if there are multiple tests in a single file, then the 2nd test needs to drop the table created by the 1st test? 

That `drop` statement should be ok since it will drop from the `install-check` schema and not from the `madlib` schema since we've already created a table in the `install-check` schema and it is first in the search path.  

We could also schema qualify the first drop statement with the install-check schema name. This name can be put into a variable by the preprocessor and then used in the install-check. This would be less inconvenient since it will only be for the drop statements (and potentially only for the first drop statement).

jingyimei:
@iyerr3 The install-check schema is by module, and we may have multiple files for one module. 

For your proposed solution, what do you expect in the single source of table creation? Do you mean audit all the table creation and try to combine all the data into one big table and every sql just use the same table? For example, for pagerank and hits in `graph` module,  we only create one vertex and one edge table with all the special cases/data, and always use those two tables for all the sql files? Besides, how do we deal with some special test case when we only need 1 row/ few row in the table instead of a bigger table? Also in some DEV check we assert a specific sql returns a specific number of rows, and those assertion will fail because of the change of input table.

The other concern is that this solution requires lots of audition and changes in every sql file. Developers need to understand how to add new row in the single table creation file and how sql files under the same module work together with each other. For example, if someone wants to add a test case in HITS under graph, instead of adding a new case in hits sql file, he needs to go to another file and add table, and go back to hits sql file and add test case. It sounds a bit confusing and inconvenient to me.

iyerr3:
I'm getting confused here, so maybe an example would help. 

My understanding of @orhankislal 's comment was that say `page_rank` and `bfs` both use the same table called "vertex" then we will have to drop this table at the beginning of each of the test files. This may then collide with a user-created "vertex" table - hence this commit. 
My suggestions to solve the problem: 
- Use a function-specific name (`"pagerank_vertex"`) instead of a generic name with (`"vertex"`) no drop by in the beginning. 
- Create a single `"vertex"` table (with no drop by statement) and use that for all graph functions
- Schema-qualify the data table with the install-check schema name. 

Please note I'm not suggesting pooling **all** install-check data into a single file (at least not as a solution to this problem).

iyerr3:
> > * Use a function-specific name (`"pagerank_vertex"`) instead of a generic name with (`"vertex"`) no drop by in the beginning.
> 
> This seems doesn't solve the potential conflict with user created table.  

It does with the original suggestion of not dropping a newly created table the first time. That suggestion led to this discussion about same table name being reused and the need to `drop table` before `create table` (which should not actually be a problem - see end of comment). 

>Besides, in some IC/DC file we drop tables in between and create table with the same name again. In this case the second drop/creation will fail if user already created it.

No; the table *created* in install-check will be in the install-check schema and the subsequent drop will look inside that schema. The original issue is due to a `drop table` statement on a table that was *not created in the file*. 

> > * Create a single `"vertex"` table (with no drop by statement) and use that for all graph functions
>
> This may cause some assertion based on returning number of records fail as I mentioned in the previous comment, and also needs effort of audition and combining test data.

If the data required in a table is different for different tests then we would benefit by giving it different names. This is a problem mostly in graphs where "vertex" and "edge" are re-used. These names are already vague and should ideally be renamed. 
 
> > * Schema-qualify the data table with the install-check schema name.
>
> (not sure if I totally understand this) install-check schema is based on module, not algo, for example pagerank and hits and sssp use the same IC schema, and there will be issue between running two sql files for droping/creating same table.

Schema-qualifying was an alternative suggestion to avoid looking into the `madlib` schema. With the schema-qualification, we can drop the table just like we're doing now (so no conflicts between test files). 

**Also note:** `madpack` is supposed to drop the file-specific schema after executing each file (see function [_execute_per_module_install_dev_check_algo](https://github.com/apache/madlib/blob/03e82bc50494367c6b65dced66cf0042b3060042/src/madpack/madpack.py#L768)). Hence, common table names in independent tests within same module are not supposed to conflict with each other (if you've seen this happen then it requires investigation). 

Finally, a guideline MADlib followed a few years ago was to not issue drop table statements in IC since all created tables are in a unique schema and will be dropped at end of test execution (I plead guilty of ignoring this guideline). This requires the names of data and output tables within a file to be unique - hopefully, we give such tables meaningful names so that we can identify the cause of a failure from the name. 

**In summary**: I suggest we don't circumvent the original issue, rather fix it by giving meaningful names and avoiding `drop table` statements in IC/DC.

orhankislal:
> Also note: madpack is supposed to drop the file-specific schema after executing each file (see function _execute_per_module_install_dev_check_algo). Hence, common table names in independent tests within same module are not supposed to conflict with each other (if you've seen this happen then it requires investigation).

Oh, I see. The naming convention is somewhat strange. Under the modules folder, we have a bunch of folders (graph, etc.) but they are not actually modules, the individual sql files are. The madpack code uses the variable `module` for the folder name which further muddles the naming. This means `madlib_installcheck_graph` schema will be created and dropped for each module in graph. We might want to change it to reflect the actual module name. 

I think the reused name issue might be more widespread than we think. I am pretty sure `abalone` and `houses` datasets are used in multiple modules.

I think removing the `DROP TABLE` statements might work as @iyerr3 suggested. I'll keep the PR open for now to keep the conversation open and start working on a different branch.

orhankislal:
Another alternative is to integrate the madlib deployment into the IC/DC. What I mean is similar to how PostGIS runs its unit tests. IC/DC creates a temporary database/schema, deploys the MADlib over there, runs the tests as usual and then removes the temporary database/schema. This will inevitably increase the running time of IC/DC but I believe it will be more stable. Since we assume that the user might be using the madlib deployment schema, it is also possible that they drop and/or recreate UDTs, UDFs and UDAs. Our IC/DC does not account for a case like that and will probably fail. 

It will also mean that a user can run IC/DC before they deploy it to their target database. I would assume most users are already using a similar workflow (temp database -> deploy MADlib -> run IC -> deploy MADlib on actual target)

asfgit:
https://builds.apache.org/job/madlib-pr-build/699/

PR #328, opened 2018-10-01
Graph: Add id of nodes with 0 in-degree

iyerr3:
JIRA: MADLIB-1279

IDs of nodes with 0 in-degree were not showing in the output of
`in_out_degrees` since the output table was a result of a full outer
join where the id was obtained from only one side.

This has been fixed by checking for NULL values (using coalesce) and the
result from other side is obtained if the ID is missing on primary side.

Closes #328


asfgit:
https://builds.apache.org/job/madlib-pr-build/694/

asfgit:
https://builds.apache.org/job/madlib-pr-build/695/

PR #330, opened 2018-10-03
Margins: Copy summary table instead of renaming

orhankislal:
JIRA: MADLIB-1274

Margins summary table gets dropped since its schema remains pg_temp.
This commit fixed the issue by copying the contents instead of renaming.

### DIFF
--- a/src/ports/postgres/modules/regress/marginal.py_in
+++ b/src/ports/postgres/modules/regress/marginal.py_in
@@ -15,7 +15,6 @@ from utilities.utilities import _assert
 from utilities.utilities import py_list_to_sql_string
 
 from utilities.validate_args import table_exists
-from utilities.validate_args import rename_table
 from utilities.validate_args import columns_exist_in_table
 from utilities.validate_args import table_is_empty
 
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/regress/marginal.py_in
```diff
@@ -522,9 +520,8 @@ def margins_mlogregr(schema_madlib, source_table, out_table,
             """.format(**all_arguments))
 
         # Rename the output summary table
-        rename_table(schema_madlib,
-                     "{old}_summary".format(old=mlog_out_table),
-                     "{new}_summary".format(new=out_table))
+        plpy.execute("""CREATE TABLE {out_table}_summary AS
```
kaknikhil: we should ideally drop the `mlog_out_table` table even if it is in the pg_temp schema. Otherwise it will be always be present in the current session.

kaknikhil: can we add a dev-check test to assert that the output summary table should exist ?

kaknikhil:
Also the jira no should be MADLIB-1276 instead of MADLIB-1274.

asfgit:
https://builds.apache.org/job/madlib-pr-build/700/

PR #331, opened 2018-10-09
Build: Include preflight and postflight scripts for mac

kaknikhil:
Commit 441f16bd55d2a26e4dd59df6129c6092f099cbca introduced a bug where
the preflight and postflight scripts for mac were not getting included
in the .dmg file. This commit adds an if check for the APPLE platform to
include these scripts necessary for dmg installation.


asfgit:
https://builds.apache.org/job/madlib-pr-build/702/

orhankislal:
Good catch +1

PR #337, opened 2018-10-26
Madpack: Add UDO and UDOC automation

orhankislal:
JIRA: MADLIB-1281

- Add scripts for detecting changed/dropped UDOs and UDOCs.
- Expand the create_changelist.py file to consume these scripts and
create changelists with these fields filled if necessary.
- Fix the update_util.py to use the correct dictionary key.
- Add drop operator class command to the svac.sql_in to make sure the
old class is removed before creating the updated one.


asfgit:
https://builds.apache.org/job/madlib-pr-build/707/

src/madpack/create_changelist.py
```diff
@@ -237,6 +325,13 @@
     print "Something went wrong! The changelist might be wrong/corrupted."
     raise
 finally:
-    os.system("rm -f /tmp/madlib_tmp_nm.txt /tmp/madlib_tmp_udf.txt "
-              "/tmp/madlib_tmp_udt.txt /tmp/madlib_tmp_cl.yaml "
```
jingyimei: what is this madlib_tmp_cl.yaml file here? why we don't remove it now

src/madpack/diff_udo.sql
```diff
@@ -0,0 +1,81 @@
...
+            FROM (
+                SELECT n.nspname AS "Schema",
+                       o.oprname AS name,
+                       filter_schema(o.oprcode::text, '{schema_name}') AS oprcode,
+                       alter_schema(pg_catalog.format_type(o.oprleft, NULL), '{schema_name}') AS oprleft,
+                       alter_schema(pg_catalog.format_type(o.oprright, NULL), '{schema_name}') AS oprright,
+                       alter_schema(pg_catalog.format_type(o.oprresult, NULL), '{schema_name}') AS rettype
+                FROM pg_catalog.pg_operator o
+                    LEFT JOIN pg_catalog.pg_namespace n ON n.oid = o.oprnamespace
+                WHERE n.nspname OPERATOR(pg_catalog.~) '^({schema_name})$'
```
jingyimei: can you please explain this where clause a bit? I am a bit confused about the 'OPERATOR' part

src/madpack/create_changelist.py
orhankislal: Nice catch, it should still be removed.

src/madpack/diff_udo.sql
orhankislal: I use the `\do madlib.*` command of `psql` as a basis. The corresponding query (you can get this if you start with `psql -E`) uses this particular phrase to get all of the operators of a particular schema. 
Basically, this regex looks at the schema name(n.nspname) and filters that don't start (^) and end ($) with madlib schema name.

asfgit:
https://builds.apache.org/job/madlib-pr-build/708/

src/madpack/diff_udo.sql
jingyimei: Got it. Thanks for the explanation!

PR #339, opened 2018-11-26
Build: Add PG11 Support

orhankislal:
JIRA: MADLIB-1283

PG11 support required a number of minor changes in the code.
- Change TRUE/FALSE to true/false
- Use TupleDescAttr function instead of direct access.
- Use prokind column instead of proisagg.

We also added a function to check if the PG version is earlier than 11
as well as the necessary cmake files.

### DIFF
--- a/src/ports/postgres/modules/kmeans/kmeans.sql_in
+++ b/src/ports/postgres/modules/kmeans/kmeans.sql_in
@@ -767,27 +767,16 @@ BEGIN
     proc_fn_dist := fn_dist
         || '(DOUBLE PRECISION[], DOUBLE PRECISION[])';
 
-    -- Handle PG11 pg_proc table changes
-    IF (SELECT MADLIB_SCHEMA.is_less_than_pg11() = TRUE) THEN
-      IF (SELECT prorettype != 'DOUBLE PRECISION'::regtype OR proisagg = TRUE
-          FROM pg_proc WHERE oid = proc_fn_dist) THEN
-          RAISE EXCEPTION 'Kmeans error: Distance function has wrong signature or is not a simple function.';
-      END IF;
-      proc_agg_centroid := agg_centroid || '(DOUBLE PRECISION[])';
-      IF (SELECT prorettype != 'DOUBLE PRECISION[]'::regtype OR proisagg = FALSE
-          FROM pg_proc WHERE oid = proc_agg_centroid) THEN
-          RAISE EXCEPTION 'Kmeans error: Mean aggregate has wrong signature or is not an aggregate.';
-      END IF;
-    ELSE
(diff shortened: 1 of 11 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/712/

src/ports/postgres/modules/utilities/utilities.py_in
```diff
@@ -407,6 +407,15 @@ def set_client_min_messages(new_level):
     return old_msg_level
 # -------------------------------------------------------------------------
 
+def is_less_than_pg11(schema_madlib, **kwargs):
```
kaknikhil: I noticed that we have another function called `is_less_than_pg90` which has the same code as this function. We probably don't need the pg90 function anymore but it would be nice if the `is_less_than_pg11`  function can be generalized to take in a major postgres version and then compare the current version with the passed major version.

src/ports/postgres/modules/kmeans/kmeans.sql_in
```diff
@@ -766,15 +766,30 @@ BEGIN
...
-    IF (SELECT prorettype != 'DOUBLE PRECISION'::regtype OR proisagg = TRUE
-        FROM pg_proc WHERE oid = proc_fn_dist) THEN
-        RAISE EXCEPTION 'Kmeans error: Distance function has wrong signature or is not a simple function.';
-    END IF;
-    proc_agg_centroid := agg_centroid || '(DOUBLE PRECISION[])';
-    IF (SELECT prorettype != 'DOUBLE PRECISION[]'::regtype OR proisagg = FALSE
-        FROM pg_proc WHERE oid = proc_agg_centroid) THEN
-        RAISE EXCEPTION 'Kmeans error: Mean aggregate has wrong signature or is not an aggregate.';
+
+    -- Handle PG11 pg_proc table changes
```
kaknikhil: We can write a sql function called `is_valid_distance_function()` that can be called from either python or sql. This function will abstract out the pg11 changes and we would just call this function from kmeans.sql_in and knn.py_in.

asfgit:
https://builds.apache.org/job/madlib-pr-build/714/

LaoTsing:
hello! i use PG12 and will try to modify python script but get error: SQLCommon.m4 not found - please add PG12 support

fmcquillan99:
I added a JIRA for this
https://issues.apache.org/jira/browse/MADLIB-1391

We will get to it in the next release after 1.17

kaknikhil:
@LaoTsing You can either wait until the next release after 1.17 or the alternative is to make changes to the code to add support for pg12. It's best to wait for the official release which will be well tested. But if you can't wait and need to get it done before that, then you will have to make a few code changes including adding a couple of cmake files. 
If you are interested in making the changes yourself, let us know and we can guide you through the process.

PR #350, opened 2019-02-05
Use 64 bit flag for compiling MADlib with greenplum on macOS.

kaknikhil:
We used to assume that gpdb on mac is 32 bit but that is not the case
anymore.


asfgit:
https://builds.apache.org/job/madlib-pr-build/744/

PR #356, opened 2019-03-14
Keras model arch table helper functions for keras_fit()

reductionista:
This PR introduces two new helper functions, `load_keras_model()` and `delete_keras_model()`.  It is a part of the new deep learning functionality being added to madlib.

These functions will help prepare the input to the `keras_fit()` function being introduced in PR [#355](https://github.com/apache/madlib/pull/355).

`keras_fit()` takes a model_arch_table param and a model_id param.  This refers to a row in a "keras model arch table" which has to have a particular format.  These functions will help the user to create and manage that table, without worrying about the details of the format.

Each row of the keras model arch table represents a keras model architecture, and is identified by a model_id.  `load_keras_model()` will add a new model architecture to the table, creating the table if it doesn't exist and then inserting the appropriate row into it.  The newly assigned model_id will be printed for the user, who can then pass that as the model_id parameter to `keras_fit()`.

The model architectures themselves are expected to be in JSON format, which can be achieved by calling `model.to_json()` on any keras model after it is created (in python).

The keras model arch table also has a column to store weights.  This is initialized to `null` by `load_keras_model()` but we plan to add another helper function in the near future to be able to set the weights (useful for doing a warm start, instead of letting keras randomly initialize the weights when training begins).

The keras model arch table itself will be automatically dropped by `delete_keras_model()` after all of the models in it are deleted and the table is empty.

To help avoid mistakes, if either of these functions is run on a table that doesn't match the expected format, and exception will be thrown and the table will be left untouched.


asfgit:
https://builds.apache.org/job/madlib-pr-build/773/

src/ports/postgres/modules/convex/test/keras_model_arch_table.ic.sql_in
```diff
@@ -0,0 +1,31 @@
...
+/* -----------------------------------------------------------------------------
+ * Test Keras Model Arch Table helper functions
+ * -------------------------------------------------------------------------- */
+
+SELECT keras_create_model('test239587_keras_model_arch_table','{"a" : 1, "b" : 2, "c" : [4,5,6] }');
+SELECT keras_create_model('test239587_keras_model_arch_table','{"config" : [1,2,3]}');
+SELECT keras_create_model('test239587_keras_model_arch_table','{"config" : [8,4,0]}');
+SELECT keras_delete_model('test239587_keras_model_arch_table',2);
+SELECT keras_delete_model('test239587_keras_model_arch_table',3);
+SELECT keras_delete_model('test239587_keras_model_arch_table',1);
```
njayaram2: Can we please end the file with a new line.

src/ports/postgres/modules/convex/test/keras_model_arch_table.ic.sql_in
```diff
@@ -0,0 +1,31 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ * ----------------------------------------------------------------------- */
+
+/* -----------------------------------------------------------------------------
+ * Test Keras Model Arch Table helper functions
+ * -------------------------------------------------------------------------- */
+
+SELECT keras_create_model('test239587_keras_model_arch_table','{"a" : 1, "b" : 2, "c" : [4,5,6] }');
```
njayaram2: The name of the function created to do this is `load_keras_model`, but the tests call `keras_create_model`. This would fail.

src/ports/postgres/modules/convex/test/keras_model_arch_table.sql_in
```diff
@@ -0,0 +1,124 @@
...
+ * ----------------------------------------------------------------------- */
+
+/* -----------------------------------------------------------------------------
+ * Test Keras Model Arch Table helper functions
+ * -------------------------------------------------------------------------- */
+
+
+/* Test successful model creation where no table exists */
+DROP TABLE IF EXISTS test_keras_model_arch_table;
+SELECT keras_create_model('test_keras_model_arch_table', '{"a" : 1, "b" : 2, "c" : [4,5,6] }');
```
njayaram2: The name of the function created to do this is `load_keras_model`, but the tests call `keras_create_model`. This would fail. This is the case in all test cases I guess.

src/ports/postgres/modules/convex/test/keras_model_arch_table.sql_in
```diff
@@ -0,0 +1,124 @@
...
+    FROM test_keras_model_arch_table WHERE model_id = 2;
+SELECT assert(model_arch->'config'->>0 = '8', 'Cannot parse model_arch json in model table.')
+    FROM test_keras_model_arch_table WHERE model_id = 3;
+SELECT assert(model_arch->'config'->>1 = '4', 'Cannot parse model_arch json in model table.')
+    FROM test_keras_model_arch_table WHERE model_id = 3;
+SELECT assert(model_arch->'config'->>2 = '0', 'Cannot parse model_arch json in model table.')
+    FROM test_keras_model_arch_table WHERE model_id = 3;
+
+/* Test deletion where valid table exists */
+SELECT keras_delete_model('test_keras_model_arch_table', 2);
```
njayaram2: The name of the function created to do this is `delete_keras_model`, but the tests call `keras_delete_model`. This would fail. This is the case in other similar test cases too I guess.

src/ports/postgres/modules/convex/keras_model_arch_table.py_in
```diff
@@ -0,0 +1,133 @@
...
+
+    sql = """
+           DELETE FROM {model_arch_table} WHERE model_id={model_id}
+          """.format(model_arch_table=model_arch_table, model_id=model_id)
+    res = _execute(sql)
+
+    if res.nrows() > 0:
+        plpy.info("Model id {0} has been deleted from {1}.".format(model_id, model_arch_table))
+    else:
+        plpy.info("Model id {0} not found".format(model_id))
```
njayaram2: Should this be `plpy.error()` instead of `info`?

asfgit:
https://builds.apache.org/job/madlib-pr-build/775/

src/ports/postgres/modules/convex/keras_model_arch_table.py_in
```diff
@@ -0,0 +1,132 @@
...
+
+    sql = "SELECT model_id FROM {model_arch_table}".format(model_arch_table=model_arch_table)
+    res = _execute(sql)
+    if not res or len(res) == 0:
```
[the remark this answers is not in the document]
reductionista: Good point, that should cover both cases.  I think this is in a couple spots.

src/ports/postgres/modules/convex/keras_model_arch_table.py_in
reductionista: I don't feel strongly either way, but we should discuss as a group and decide.  I think the question boils down to whether we want it to act more like "DROP" or "DROP IF EXISTS", both of which can be useful at times.

reductionista: Oh, actually maybe plpy.warn()?  I think I like that more than error or info... warn the user, but don't throw an exception.

src/ports/postgres/modules/convex/test/keras_model_arch_table.sql_in
```diff
@@ -0,0 +1,124 @@
...
+BEGIN
+    BEGIN
+        EXECUTE stmt;
+	EXCEPTION
+        WHEN OTHERS THEN
+            RETURN 1;
+    END;
+    RETURN 0;
+END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
reductionista: Agreed, it should go somewhere where all tests can access it.

src/ports/postgres/modules/convex/keras_model_arch_table.py_in
```diff
@@ -0,0 +1,132 @@
...
+        plpy.info("Removing empty keras model arch table {model_arch_table}".format(model_arch_table=model_arch_table))
+        sql = "DROP TABLE {model_arch_table}".format(model_arch_table=model_arch_table)
```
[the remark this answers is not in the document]
reductionista: Yes, I think I agree.

Current behavior is that the user would see an exception and an error message saying that the table cannot be dropped due to the view.  Then the whole transaction will be rolled back, so the model they wanted deleted from the table will not be deleted.

If we handle the exception, we could replace it with our own error message (which would probably be nearly identical), but avoid throwing an exception so that the model does get removed from the table.

I'm not entirely sure that the second behavior is more desirable, but I think so.  The only weird part of it is that it leaves the table in a state it wouldn't otherwise be able to get into.  In all other cases, either there is no table or there is at least one model in the table; while in this case it would leave an empty table laying around.   But it shouldn't hurt anything.

asfgit:
https://builds.apache.org/job/madlib-pr-build/776/

asfgit:
https://builds.apache.org/job/madlib-pr-build/779/

PR #360, opened 2019-03-28
Deep Learning: Add support for one-hot encoded dep var

njayaram2:
The current madlib_keras_fit function for DL assumes dependent variable is not
one-hot encoded. But fit uses data obtained after running
minibatch_preprocessor_dl that returns a 1-hot encoded array for each dependent
var (https://issues.apache.org/jira/browse/MADLIB-1303).

This commit updates madlib_keras_fit to expect a 1-hot encoded dependent var to
train the model. We also create a new column named class_values in the model
summary table that is passed along from the input summary table (which is
mini-batched).  Predict then uses class_values to find the exact class label
for test data. class_values can be NULL in case where user had already 1-hot
encoded the data before calling minibatch_preprocessor_dl. In that case, we
return the index of the expected class label.


asfgit:
https://builds.apache.org/job/madlib-pr-build/797/

asfgit:
https://builds.apache.org/job/madlib-pr-build/798/

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
```diff
@@ -174,21 +182,129 @@ def get_data_as_np_array(table_name, y, x, input_shape, num_classes):
...
+        self.independent_varname = independent_varname
+        self.num_iterations = num_iterations
+        self.source_summary_table = None
+        if self.source_table:
+            self.source_summary_table = add_postfix(
+                self.source_table, "_summary")
+        if self.output_model_table:
+            self.output_summary_model_table = add_postfix(
+                self.output_model_table, "_summary")
+        self.class_values_colname = CLASS_VALUES_COLNAME
```
kaknikhil: Why do we need another variable for this ? Why can't we just use CLASS_VALUES_COLNAME?

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -75,9 +89,24 @@ def internal_keras_predict(x_test, model_arch, model_data, input_shape, compile_
...
-    res = model.predict_classes(x_test)
-    return res
+    proba_argmax = model.predict_classes(x_test)
+    # proba_argmax is a list with exactly one element in it. That element
+    # refers to the index containing the largest probability value in the
+    # output of Keras' predict function.
+    return _get_class_label(class_values, proba_argmax[0])
+
+def _get_class_label(class_values, class_index):
+    if class_values:
```
kaknikhil: Might be a good idea to add docstring for this function. We should also add a comment explaining the sorting order of class_values

kaknikhil: Also maybe consider adding unit tests for this function

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -296,16 +217,18 @@ def fit(schema_madlib, source_table, model, dependent_varname,
         $24 AS accuracy_validation,
         $25 AS loss_validation,
         $26 AS accuracy_iter_validation,
-        $27 AS loss_iter_validation
+        $27 AS loss_iter_validation,
+        $28 AS class_values
```
kaknikhil: can we use the `CLASS_VALUES_COLNAME` variable here instead of hardcoding it ?

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
```diff
@@ -174,21 +182,129 @@ def get_data_as_np_array(table_name, y, x, input_shape, num_classes):
...
+                # meaningful error msg.
+                input_shape_from_table = [result["n_{0}".format(i)]
+                    for i in range(len(input_shape))]
+                plpy.error("model_keras error: Input shape {0} in the model" \
+                    " architecture does not match the input shape {1} of column" \
+                    " {2} in table {3}.".format(
+                        input_shape, input_shape_from_table,
+                        self.independent_varname, table))
+
+def get_class_values_and_type(class_values_colname, source_summary_table):
```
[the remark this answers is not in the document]
njayaram2: Nothing in `madlib_keras_helper.py_in` has keras related code (in the sense that nothing imports and uses anything related to keras). All of keras related helper functions are in `madlib_keras_wrapper.py_in`. So I think this is the right place for this function.

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -31,103 +31,21 @@ from keras.models import *
 from keras.optimizers import *
 from keras.regularizers import *
 
-from madlib_keras_helper import KerasWeightsSerializer
+from madlib_keras_helper import FitInputValidator
```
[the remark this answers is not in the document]
njayaram2: I guess this is best addressed when the future commit happens. But, I am not sure why we need a new file called `madlib_keras_serializer.py_in`. I think it's fine as is, as a class in the helper function.

If we have another class to validate predict, we can have a separate class for that. That would also have validate input code, and it might be a good idea to containerize these functions for fit and predict inside their own classes. Moreover, we also have `validate_input_shapes` as one of the functions in `class FitInputValidator`.

asfgit:
https://builds.apache.org/job/madlib-pr-build/801/

PR #361, opened 2019-03-30
Minibatch Preprocessor DL: Add optional num_classes param.

njayaram2:
The current `minibatch_preprocessor_dl()` module looks at the input
table to find the number of distinct categories (class values) for the
dependent variable, and uses that number as the size of the
one-hot-encoded array. This could lead to a failure in madlib_keras fit
function if the `num_classes` defined in the architecture is a number
greater/different than the size of the one hot encoded array.

This commit adds two functionalities:
1) A new optional parameter to `minibatch_preprocessor_dl()` that will
be used to determine the length of the 1-hot encoded vector for the
dependent var. If the param is set to NULL, the length will be equal to
the number of distinct class values found in the dataset, else
num_classes must be greater than equal to the number of distinct class
values.
The `class_values` column in the summary table contains an array of
class values associated with the 1-hot encoded vector. That will have
NULL as the value for class values that we don't find any representation
for in the dataset.
2) We now support NULL as a valid class value for dependent variable.

### DIFF
--- a/src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
+++ b/src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
@@ -86,7 +86,8 @@ class MiniBatchPreProcessor:
         self.to_one_hot_encode = self.should_one_hot_encode(one_hot_encode_int_dep_var)
         if self.to_one_hot_encode:
             self.dependent_levels = get_distinct_col_levels(
-            self.source_table, self.dependent_varname, self.dependent_vartype)
+                self.source_table, self.dependent_varname,
+                self.dependent_vartype)
         else:
             self.dependent_levels = None
         self._validate_minibatch_preprocessor_params()
@@ -343,10 +344,10 @@ class MiniBatchPreProcessor:
         plpy.execute(query)
 
 
-class MiniBatchPreProcessorDL(MiniBatchPreProcessor):
+class MiniBatchPreProcessorDL():
     def __init__(self, schema_madlib, source_table, output_table,
                  dependent_varname, independent_varname, buffer_size,
-                 normalizing_const=1.0, **kwargs):
+                 normalizing_const, num_classes, **kwargs):
         self.schema_madlib = schema_madlib
         self.source_table = source_table
         self.output_table = output_table
@@ -354,6 +355,7 @@ class MiniBatchPreProcessorDL(MiniBatchPreProcessor):
         self.independent_varname = independent_varname
         self.buffer_size = buffer_size
         self.normalizing_const = normalizing_const if normalizing_const is not None else 1.0
+        self.num_classes = num_classes
         self.module_name = "minibatch_preprocessor_DL"
         self.output_summary_table = add_postfix(self.output_table, "_summary")
         self.independent_vartype = get_expr_type(self.independent_varname,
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/799/

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -363,21 +365,70 @@ class MiniBatchPreProcessorDL(MiniBatchPreProcessor):
 
         self._validate_args()
         self.num_of_buffers = self._get_num_buffers()
-        self.to_one_hot_encode = True
```
kaknikhil: I think we should document our motivation for always one hot encoding the dependent variable? I would say as a comment here in the code and also in the docs

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -762,6 +823,11 @@ class MiniBatchDocumentation:
             normalizing_const      -- DOUBLE PRECISON. Default 1.0. The
                                       normalizing constant to use for
                                       standardizing arrays in independent_varname.
+            num_classes            -- INTEGER. Default NULL. Number of class labels
+                                      to be considered for 1-hot encoding. If None,
```
kaknikhil: /s/None/NULL

src/ports/postgres/modules/utilities/minibatch_preprocessing_dl.sql_in
```diff
@@ -105,6 +106,12 @@ minibatch_preprocessor_dl( source_table,
   each value in the independent_varname array by.  For example,
   you may need to use 255 for this value if the image data is in the form 0-255.
   </dd>
+
+  <dt>num_classes (optional)</dt>
+  <dd>INTEGER, default: NULL. Number of class labels to be considered for 1-hot
+  encoding. If None, the 1-hot encoded array length will be equal to the number
```
kaknikhil: /s/None/NULL

src/ports/postgres/modules/utilities/test/minibatch_preprocessing_dl.sql_in
```diff
@@ -205,3 +210,81 @@ SELECT assert (dependent_vartype   = 'integer[]' AND
...
+  'label',
+  'x',
+  4,
+  5,
+  5 -- num_classes
+  );
+-- Test summary table if class_values has NULL as a legitimate
+-- class label, and also two other NULLs because num_classes=5
+-- but table has only 3 distinct class labels (including NULL)
+SELECT assert
```
kaknikhil: +1 for adding this test

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -363,21 +365,70 @@ class MiniBatchPreProcessorDL(MiniBatchPreProcessor):
...
+                    2. dependent_varname = ARRAY[a, b, c]
+                        transformed_value = ARRAY[a, b, c]
+            else if dep_type in ("text", "boolean"), encode:
+                    3. dependent_varname = rings (encoding)
+                        transformed_value = ARRAY[rings=1, rings=2, rings=3]
+        """
+        # Assuming the input NUMERIC[] is already one_hot_encoded,
+        # so casting to INTEGER[]
+        if is_valid_psql_type(self.dependent_vartype, NUMERIC | ONLY_ARRAY):
+            return self.dependent_varname + '::INTEGER[]'
```
kaknikhil: Consider replacing the `+` with a `.format`

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
```diff
@@ -363,21 +365,70 @@ class MiniBatchPreProcessorDL(MiniBatchPreProcessor):
...
+        # like in the generic get_one_hot_encoded_expr() defined in
+        # db_utils.py_in. We also have this optional 'num_classes' param
+        # that affects the logic of 1-hot encoding. Since this is very
+        # specific to minibatch_preprocessing_dl for now, let's keep
+        # it here instead of refactoring it out to a generic helper function.
+        one_hot_encoded_expr = ["({0}) IS NOT DISTINCT FROM {1}".format(
+            self.dependent_varname, c) for c in self.dependent_levels]
+        if self.num_classes:
+            one_hot_encoded_expr.extend(['0'
+                for i in range(self.num_classes-len(self.dependent_levels))])
```
kaknikhil: Can we create a class variable for `self.num_classes-len(self.dependent_levels)` in the constructor so that it can be reused and is easy to read ?

src/ports/postgres/modules/utilities/test/unit_tests/test_minibatch_preprocessing.py_in
```diff
@@ -462,7 +487,80 @@ class MiniBatchPreProcessingDLTestCase(unittest.TestCase):
...
+        self.module.get_expr_type = Mock(side_effect = ['integer[]', 'text'])
+        obj = self.module.MiniBatchPreProcessorDL(
+            self.default_schema_madlib,
+            self.default_source_table,
+            self.default_output_table,
+            self.default_dep_var,
+            self.default_ind_var,
+            self.default_buffer_size,
+            self.default_normalizing_const,
+            None)
```
kaknikhil: Can we add an assert here?

src/ports/postgres/modules/utilities/test/unit_tests/test_minibatch_preprocessing.py_in
```diff
@@ -462,7 +487,80 @@ class MiniBatchPreProcessingDLTestCase(unittest.TestCase):
...
+        self.module.dependent_levels = Mock(return_value = ["'a'", "'b'", "'c'"])
+        obj = self.module.MiniBatchPreProcessorDL(
+            self.default_schema_madlib,
+            self.default_source_table,
+            self.default_output_table,
+            self.default_dep_var,
+            self.default_ind_var,
+            self.default_buffer_size,
+            self.default_normalizing_const,
+            5)
```
kaknikhil: Can we add an assert here?

src/ports/postgres/modules/utilities/minibatch_preprocessing.py_in
kaknikhil: Also I think we should document the fact that our implementation of one hot encoding the dep var is consistent with the output of keras.to_categorical. IMO, we should document this in our code as well as the docs.

njayaram2: Our 1-hot encoding follows the standard one-hot encoding convention. In fact, it is different from `keras.to_categorical`. For example, if there are 3 distinct class values captured in a list `y=[10, 11, 12]`, then the 1-hot encoded vector created by`keras.to_categorical(y)` is of size 13 (largest class value + 1). If it is called with `keras.to_categorical(y, num_classes=4)`, it errors out.
The 1-hot encoding done in MADlib would create a 1-hot encoded vector of size 4 in both cases.

I would say keras' 1-hot encoding is actually not the standard way of doing it.

asfgit:
https://builds.apache.org/job/madlib-pr-build/802/

asfgit:
https://builds.apache.org/job/madlib-pr-build/803/

PR #362, opened 2019-04-03
DL: Remove num_classes param from madlib_keras_fit()

njayaram2:
The number of classes is embedded in the model architecture JSON, so it
is not necessary to explicitly pass it again as a parameter to fit. This
commit parses the model architecture JSON to automatically figure out
the num_classes value to use.

### DIFF
--- a/src/ports/postgres/modules/utilities/model_arch_info.py_in
+++ b/src/ports/postgres/modules/utilities/model_arch_info.py_in
@@ -21,69 +21,45 @@ m4_changequote(`<!', `!>')
 
 import sys
 import json
+import plpy
 
-def get_layers(arch):
-    d = json.loads(arch)
+def _get_layers(model_arch):
+    d = json.loads(model_arch)
     config = d['config']
     if type(config) == list:
-        return config  # In keras 1.x, all models are sequential
+        return config  # In keras 2.1.x, all models are sequential
     elif type(config) == dict and 'layers' in config:
         layers = config['layers']
         if type(layers) == list:
             return config['layers']  # In keras 2.x, only sequential models are supported
-    plpy.error('Unable to read input_shape from keras model arch.  Note: only sequential keras models are supported.')
-    return None
+    plpy.error("Unable to read model architecture JSON.")
 
-def get_input_shape(arch):
-    layers = get_layers(arch)
-    return layers[0]['config']['batch_input_shape'][1:]
+def get_input_shape(model_arch):
+    arch_layers = _get_layers(model_arch)
+    if 'batch_input_shape' in arch_layers[0]['config']:
+        return arch_layers[0]['config']['batch_input_shape'][1:]
+    plpy.error('Unable to get input shape from model architecture.')
 
-def print_model_arch_layers(arch):
-    layers = get_layers(arch)
+def get_num_classes(model_arch):
+    arch_layers = _get_layers(model_arch)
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/804/

src/ports/postgres/modules/utilities/model_arch_info.py_in
```diff
@@ -21,69 +21,45 @@ m4_changequote(`<!', `!>')
...
-    layers = get_layers(arch)
-    print("\nInput shape:")
-    print(layers[0]['config']['batch_input_shape'][1:])
-
-def print_required_imports(arch):
-    layers = get_layers(arch)
-    class_names = set(layer['class_name'] for layer in layers )
-    print("\nRequired imports:")
-    for module in class_names:
-        print("import {}".module)
```
reductionista: Some version of this function might be good to leave in.  It finds all the layer names and builds import statements out of them.  This would be useful if we ever decided to stop doing `import *` on everything in keras.  We already know which optimizer the user is using, and Sequential is the only model we support.  So I think the only left we'd need to do something more for is the regularizers.

src/ports/postgres/modules/utilities/model_arch_info.py_in
```diff
@@ -21,69 +21,45 @@ m4_changequote(`<!', `!>')
...
     first = True
-    for layer in layers:
+    for layer in arch_layers:
         if first:
             first = False
         else:
-            print("   |")
-            print("   V")
+            layers = "{0}   |\n".format(layers)
+            layers = "{0}   V\n".format(layers)
```
reductionista: A cleaner way to write this is:
```
layers += '    |\n'
layers += '   V' 
```

src/ports/postgres/modules/utilities/model_arch_info.py_in
```diff
@@ -21,69 +21,45 @@ m4_changequote(`<!', `!>')
...
         class_name = layer['class_name']
         config = layer['config']
         if class_name == 'Dense':
-            print("{0}[{1}]".class_name)
+            layers = "{0}{1}[{2}]\n".format(layers, config, class_name)
```
reductionista: This prints out a bunch of json.  This line looks like it must have already been broken before, but I think it should be:
```
layers += '{0}[{1}]'.format(layer['units'],class_name)
```

reductionista: (prints eg. Dense[365] for fully-connected layer with 365 outputs)

src/ports/postgres/modules/utilities/model_arch_info.py_in
njayaram2: I am not sure how this will be used. If we replace `import *` by specific imports, it will the exact optimizers and layers we would support. So we should have those specific imports in our code and document exactly what layers/optimizers are supported by MADlib. 
The `madlib_keras_fit` module should fail if the model_arch has layers/optimizers etc that are not explicitly supported by MADlib. This function anyway should not be used to add new import statements and execute them on the fly. Is that your intent, or did you just want to have it around for debugging purposes?

asfgit:
https://builds.apache.org/job/madlib-pr-build/807/

reductionista:
Approved to merge!

PR #363, opened 2019-04-03
Add tests for madlib_keras_predict

reductionista:
dev-check tests:
- Validate table format of predict output
- Test that predicted values are in class value set
- Test that predictions are accurate
- Test to make sure we error out on batched input

Unit tests:
- Only _get_class_label for now.  (Tests that correct
  class value returned, plus plpy.error on invalid index.)

Other changes:
- Add normalization of training data at beginning of dev-check
  test file.  (Currently in order for predict to work, image data
  must be divided by 255 before training, to match what's done
  inside the predict function.)

- Add additional error-checking in _get_class_label so that
  unit test passes.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -108,7 +108,7 @@ def _get_class_label(class_values, class_index):
         scalar. If class_values is None, returns class_index, else returns
         class_values[class_index].
     """
-    if class_values is None:
+    if not class_values:
         return class_index
     elif class_index != int(class_index):
         plpy.error("Invalid class index {0} returned from Keras predict. "\
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -108,12 +108,15 @@ def _get_class_label(class_values, class_index):
         scalar. If class_values is None, returns class_index, else returns
         class_values[class_index].
     """
-    if class_values:
-        if class_index < len(class_values):
-            return class_values[class_index]
-        else:
-            plpy.error("Invalid class index {0} returned from Keras predict. "\
-                "Index value must be less than {1}".format(
+    if class_values is None:
+        return class_index
+    elif class_index != int(class_index):
+        plpy.error("Invalid class index {0} returned from Keras predict. "\
+                   "Index value must be an integer".format(
+                    class_index))
+    elif class_index < 0 or class_index >= len(class_values):
+        plpy.error("Invalid class index {0} returned from Keras predict. "\
+                   "Index value must be less than {1}".format(
                     class_index, len(class_values)))
     else:
-        return class_index
+        return class_values[class_index]
--- a/src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
+++ b/src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
@@ -61,10 +61,10 @@ ALTER TABLE model_arch RENAME model_id TO id;
 -- Please do not break up the compile_params string
 -- It might break the assertion
 
-DROP TABLE IF EXISTS keras_out, keras_out_summary;
+DROP TABLE IF EXISTS keras_saved_out, keras_saved_out_summary;
 SELECT madlib_keras_fit(
     'cifar_10_sample_batched',
-    'keras_out',
+    'keras_saved_out',
     'dependent_var',
     'independent_var',
     'model_arch',
@@ -83,7 +83,7 @@ SELECT assert(
         end_training_time > start_training_time AND
         source_table = 'cifar_10_sample_batched' AND
         validation_table = 'cifar_10_sample_batched' AND
-        model = 'keras_out' AND
+        model = 'keras_saved_out' AND
         dependent_varname = 'dependent_var' AND
         independent_varname = 'independent_var' AND
         name is NULL AND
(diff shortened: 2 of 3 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/805/

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
```diff
@@ -173,3 +172,36 @@ SELECT madlib_keras_predict(
...
+select assert(trap_error($$madlib_keras_predict(
+    'keras_out',
+    'cifar_10_sample_batched',
+    'id',
+    'model_arch',
+    1,
+    'x',
+    $$ optimizer=SGD(lr=0.01, decay=1e-6, nesterov=True), loss='categorical_crossentropy', metrics=['accuracy']$$::text,
+    'cifar10_predict');$$) = 1,
+    'Passing batched image table to predict should error out.');
```
njayaram2: end with a new line please.

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
```diff
@@ -173,3 +172,36 @@ SELECT madlib_keras_predict(
...
+SELECT assert(COUNT(*) = 2, "Output table of madlib_keras_predict should have two rows") FROM cifar10_predict;
+
+-- First test that all values are in set of class values; if this breaks, it's definitely a problem.
+SELECT assert(prediction in (0,1),'Predicted value not in set of defined class values for model') FROM cifar10_predict;
+
+-- Then test that each of the two images is correctly predicted.  If this breaks, it's likely a different problem, or
+   -- possibly due to the test model we chose not being trained well enough.  Based on current testing, it looks
+   -- reliable.
+SELECT assert(prediction=0,'Predicted value not in set of defined class values for model') FROM cifar10_predict WHERE id=1;
+SELECT assert(prediction=1,'Predicted value not in set of defined class values for model') FROM cifar10_predict WHERE id=2;
```
njayaram2: Might not be a great idea to assert for exact prediction values. The model learnt in this test scenario is most likely something random, and we may end up seeing different results for different runs. We can instead assert that it is one of 0 or 1, like in the test in line 190.

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
```diff
@@ -173,3 +172,36 @@ SELECT madlib_keras_predict(
...
+    FROM pg_attribute WHERE attrelid = 'cifar10_predict'::regclass
+        AND attname = 'id';
+
+SELECT assert(UPPER(atttypid::regtype::TEXT) =
+    'DOUBLE PRECISION', 'prediction column should be DOUBLE PRECISION type')
+    FROM pg_attribute WHERE attrelid = 'cifar10_predict'::regclass
+        AND attname = 'prediction';
+
+-- Validate correct number of rows returned.
+SELECT assert(COUNT(*) = 2, "Output table of madlib_keras_predict should have two rows") FROM cifar10_predict;
```
njayaram2: This assert failed with the following error on GPDB 5:
```
-- Validate correct number of rows returned.
SELECT assert(COUNT(*) = 2, "Output table of madlib_keras_predict should have two rows") FROM cifar10_predict;
psql:/tmp/madlib.GlzAcb/deep_learning/madlib_keras.sql_in.tmp:192: ERROR:  column "Output table of madlib_keras_predict should have two rows" does not exist
LINE 1: SELECT assert(COUNT(*) = 2, "Output table of madlib_keras_pr...
```

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
njayaram2: I actually tried running it multiple times, and it did have different predictions. The assert failed some times.

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -108,12 +108,15 @@ def _get_class_label(class_values, class_index):
         scalar. If class_values is None, returns class_index, else returns
         class_values[class_index].
     """
-    if class_values:
-        if class_index < len(class_values):
-            return class_values[class_index]
-        else:
-            plpy.error("Invalid class index {0} returned from Keras predict. "\
-                "Index value must be less than {1}".format(
+    if class_values is None:
```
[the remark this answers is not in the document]
reductionista: My understanding is we don't want to treat an empty list as None.  Empty list should return a range error.  If the intended behavior is that empty list should return the index, then we should change the doc string to say that (currently says None is used for that purpose).

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
reductionista: Thanks, this was due to using double quotes instead of single.

src/ports/postgres/modules/deep_learning/test/madlib_keras.sql_in
reductionista: This turned out to be a bug due to keras_out getting overwritten.  It was using the wrong output table.  As long as it uses the one from the first call to keras_fit (which runs for 3 iterations instead of just 1) the validation accuracy is always 100%.

I've fixed this now and added a test of keras_fit to ensure the accuracy is 100% after training.  Since there are only 2 training examples, if it can't classify them into 2 different classes by 3 iterations, I think that should be taken as a sign that something's broken.

asfgit:
https://builds.apache.org/job/madlib-pr-build/806/

asfgit:
https://builds.apache.org/job/madlib-pr-build/808/

asfgit:
https://builds.apache.org/job/madlib-pr-build/810/

PR #365, opened 2019-04-09
DL: Add new columns in fit output summary table.

njayaram2:
JIRA: MADLIB-1319

Columns `dependent_vartype` and `normalizing_const` are useful to have
in the model summary table since they can be used in predict.
1) For instance, `dependent_vartype` will be useful in
https://issues.apache.org/jira/browse/MADLIB-1315 while creating the
prediction output column. In case of response, the output column should
be of the same type as the dependent variable in train data.
2) While predicting, `normalizing_const` will be useful to normalize the
test data using the same normalizing constant that was used to normalize
the train data.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -29,7 +29,6 @@ import numpy as np
 
 from utilities.model_arch_info import get_input_shape
 from utilities.utilities import add_postfix
-from utilities.validate_args import get_col_value_and_type
 from utilities.validate_args import input_tbl_valid
 from utilities.validate_args import output_tbl_valid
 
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/811/

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -60,7 +61,8 @@ def predict(schema_madlib, model_table, test_table, id_col, model_arch_table,
     input_shape = get_input_shape(model_arch)
     compile_params = "$madlib$" + compile_params + "$madlib$"
     model_summary_table = add_postfix(model_table, "_summary")
-    class_values, _ = get_class_values_and_type(model_summary_table)
+    class_values, _ = get_col_value_and_type(
```
khannaekta: Instead of calling `get_col_value_and_type()` here, can we do something similar to what we have in [madlib_keras.py_in:202](https://github.com/apache/madlib/pull/365/files#diff-4d3eba8c7b20343818ac24e5730779e7R202)

src/ports/postgres/modules/utilities/validate_args.py_in
```diff
@@ -365,6 +365,24 @@ def get_cols_and_types(tbl):
     return list(zip(col_names, col_types))
 # -------------------------------------------------------------------------
 
+def get_col_value_and_type(table_name, column_name):
```
khannaekta: We could extend this function to handle multiple column_names in addition to a single column_name.

njayaram2: That's a good suggestion, but it may not be necessary at the moment. If we see this function getting used more often in the future, let's refactor it out and update it this way.

asfgit:
https://builds.apache.org/job/madlib-pr-build/812/

PR #370, opened 2019-04-16
DL: Support response and prob prediction outputs

khannaekta:
JIRA: MADLIB-1315

This PR adds a new optional parameter to madlib_keras_predict called
pred_type.
1) If pred_type='response', then the prediction output would have a new
column called estimated_COLNAME, and its value would be the class label
with highest probability.
2) If pred_type='prob', then the prediction output will have multiple
columns, one for each class label. The values in these columns would be
the probability associated with the corresponding class label.


asfgit:
https://builds.apache.org/job/madlib-pr-build/825/

asfgit:
https://builds.apache.org/job/madlib-pr-build/827/

asfgit:
https://builds.apache.org/job/madlib-pr-build/828/

asfgit:
https://builds.apache.org/job/madlib-pr-build/829/

asfgit:
https://builds.apache.org/job/madlib-pr-build/830/

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -27,107 +27,113 @@ from keras.models import *
...
-    elif class_index != int(class_index):
-        plpy.error("{0}: Invalid class index {1} returned from Keras predict."\
-            " Index value must be an integer".format(MODULE_NAME, class_index))
-    elif class_index < 0 or class_index >= len(class_values):
-        plpy.error("{0}: Invalid class index {1} returned from Keras predict."\
-            " Index value must be less than {2}".format(
-                MODULE_NAME, class_index, len(class_values)))
+    # Since the test data isn't mini-batched,
+    # we have to make sure that the test data np array has the same
+    # number of dimensions as input_shape. So we a dimension to x.
```
kaknikhil: Missing the word `add`

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -198,11 +199,11 @@ def fit(schema_madlib, source_table, model, dependent_varname,
         final_validation_loss = validation_aggregate_loss[-1]
     version = madlib_version(schema_madlib)
     class_values, class_values_type = get_col_value_and_type(
-        fit_validator.source_summary_table, CLASS_VALUES_COLNAME)
+        fit_validator.source_summary_table, CLASS_VALUES_CNAME)
```
kaknikhil: Why CNAME instead of COLNAME ? I would prefer COLNAME.

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
```diff
@@ -0,0 +1,96 @@
+# coding=utf-8
```
kaknikhil: IMO, we shouldn't name this file `madlib_keras_helper.py_in` because this will just become a catch-all for all things related to deep learning. Naming it helper makes it harder to navigate through code and maintain it in the long term. For ex the function `expand_input_dims` can be in a file called something along the lines`all_things_numpy.py_in` and this file can be called `predict_input_params.py_in`.

Having said all this, I will leave the final decision up to you to decide and even if you plan to change this we can do it in a future PR.

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -27,107 +27,113 @@ from keras.models import *
...
+        pred_col_type = dependent_vartype
+    else:
+        pred_col_name = "prob"
+        pred_col_type = 'double precision'
+
+    num_of_valid_class_values = 0
+    if class_values is not None:
+        for ele in class_values:
+            if ele is None and num_of_valid_class_values > 0:
+                break
```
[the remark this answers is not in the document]
njayaram2: Done, added examples in docstring and unit tests too.

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
njayaram2: Agree and disagree. `predict_input_params.py_in` makes sense, and I have made the necessary changes for that. But for the other function, creating a new file called `all_things_numpy` may not be a good idea since we do have a lot of functions that use numpy in a different file. So that filename will be a misnomer. If we decide to name it something else, then there will be just that one function in that file, and we will also have to create a new file for the colnames variables. I'd rather have them both together in `madlib_keras_helper.py_in`. IMO, If we add more functions in future PRs, we should create appropriate files if it makes sense, else it's ok to add it in the helper file if we cannot classify it cleanly to a separate file.

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
njayaram2: Variables for some column names were getting very big, hence `CNAME`. I don't feel strongly between CNAME or COLNAME, so have changed it to COLNAME, and it's just a couple of chars anyway, so why not! :)

asfgit:
https://builds.apache.org/job/madlib-pr-build/838/

PR #374, opened 2019-04-23
DL: Validate predict input parameters

njayaram2:
JIRA: MADLIB-1321
Add validation code for predict input parameters. This commit also
refactors the code a bit and updates relevant unit tests.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_validator.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_validator.py_in
@@ -64,7 +64,7 @@ def _validate_input_shapes(table, independent_varname, input_shape, offset):
     # same number of dimensions as the input_shape.
     result = plpy.execute(query)[0]
     _assert(len(result) == len(input_shape),
-        "model_keras error: The number of dimensions ({0}) of each image" \
+        "model_keras error: The number of dimensions ({0}) of each image"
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/844/

src/ports/postgres/modules/deep_learning/madlib_keras_validator.py_in
```diff
@@ -53,15 +178,15 @@ class FitInputValidator:
 
     def _validate_input_table(self, table):
         _assert(is_var_valid(table, self.independent_varname),
-                "model_keras error: invalid independent_varname "
-                "('{independent_varname}') for table "
+                "model_keras error: invalid independent_varname "\
```
orhankislal: Do we need these `\` characters? The string does not have the new line, thanks to the way python string concatenation works.

njayaram2: Sure, will change it, I don't feel strongly either way.
I put it in there since I thought it was clearer to read the code. `\` indicates that I am splitting a single error message across multiple lines in the code (to limit the number of chars showing up in a single line in the python file), although it's still a single line as an error message itself. Python does that implicitly as you pointed out, but I put it there to show it up explicitly in the code.

PR #375, opened 2019-04-24
DL: Add GPU support for evaluate

kaknikhil:
In our experiments on a multi host gpdb cluster, we noticed that using CPU and compiling the model everytime caused the fit query to fail during the validate code with the error `Dst tensor is not
initialized`. Adding GPU and caching the compiled model fixed this issue.

1. Add GPU support to the evaluate code called by the fit function.
2. The evaluate code previously compiled the model for each row in the validation table.
This was unnecessary and takes more resources in terms of time and memory. This commit avoids this by using SD to cache the compiled model.
3. Also refactored the code by creating a function to get the rows per segment

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
@@ -59,10 +59,6 @@ def fit(schema_madlib, source_table, model, dependent_varname,
         dependent_varname, independent_varname, num_iterations)
 
     start_training_time = datetime.datetime.now()
-
-    # Disable GPU on master
-    os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
-
     use_gpu = bool(use_gpu)
 
     # Get the serialized master model
@@ -81,6 +77,19 @@ def fit(schema_madlib, source_table, model, dependent_varname,
     num_classes = get_num_classes(model_arch)
     fit_validator.validate_input_shapes(input_shape)
     model_weights_serialized = query_result[Format.MODEL_WEIGHTS]
+
+    #TODO: Refactor the pg related logic in a future PR when we think
+    # about making the fit function easier to read and maintain.
+    if is_platform_pg():
+        set_keras_session(use_gpu)
+        gp_segment_id_col =  -1
+    else:
+        gp_segment_id_col = 'gp_segment_id'
+        # Disable GPU on master for gpdb
+        os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
+    seg_ids_train, rows_per_seg_train = get_rows_per_seg_from_db(source_table)
+    if validation_table:
+        seg_ids_val, rows_per_seg_val = get_rows_per_seg_from_db(validation_table)
 
     # Convert model from json and initialize weights
     master_model = model_from_json(model_arch)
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/846/

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -391,14 +373,17 @@ def fit_transition(state, ind_var, dep_var, current_seg_id, num_classes,
...
-
-        agg_loss /= total_buffers
-        agg_accuracy /= total_buffers
-        if not is_pg:
+    if is_platform_pg():
+        total_buffers_per_seg = total_buffers_per_seg[0]
+    else:
+        total_buffers_per_seg = total_buffers_per_seg[all_seg_ids.index(current_seg_id)]
+    if total_buffers_per_seg == 0:
+        plpy.error('total buffers is 0')
```
njayaram2: Ideally, we should not be hitting this error. So, can we have a more verbose error message, and print out the `current_seg_id` too?

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -503,6 +525,21 @@ def internal_keras_evaluate(dependent_var, independent_var, model_architecture,
     independent_var = expand_input_dims(independent_var, target_type='float32')
     dependent_var = expand_input_dims(dependent_var)
 
+
     with K.tf.device(device_name):
         res = model.evaluate(independent_var, dependent_var)
+    if current_seg == -1:
```
kaknikhil: We should use `is_platform_pg()` instead of `current_seg` for this if check. It reflects the true intent of the if check

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
```diff
@@ -18,6 +18,7 @@
 # under the License.
 
 import numpy as np
+import plpy
```
njayaram2: We are not using `plpy` here, may be we should add it when we have use for it.

asfgit:
https://builds.apache.org/job/madlib-pr-build/850/

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -84,6 +80,18 @@ def fit(schema_madlib, source_table, model, dependent_varname,
...
     model_weights_serialized = query_result[Format.MODEL_WEIGHTS]
 
+    #TODO: Refactor the pg related logic in a future PR when we think
+    # about making the fit function easier to read and maintain.
+    if is_platform_pg():
+        set_keras_session(use_gpu)
+    else:
+        # Disable GPU on master for gpdb
+        os.environ['CUDA_VISIBLE_DEVICES'] = '-1'
+    gp_segment_id_col =  -1 if is_platform_pg() else 'gp_segment_id'
```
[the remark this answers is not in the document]
kaknikhil: done

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
kaknikhil: plpy.info prints the segment no along with the error message. For the message itself, I decided to make it better it in a future PR.

src/ports/postgres/modules/deep_learning/madlib_keras_helper.py_in
kaknikhil: done

asfgit:
https://builds.apache.org/job/madlib-pr-build/852/

PR #377, opened 2019-04-25
 DL: Do not compile params in predict and add gpu support

njayaram2:
JIRA: MADLIB-1330
Do not compile params in predict, but instead directly get the model
weights and architecture and use it for prediction. Compiling params
during predict is not necessary from Keras 1.0.3.

Add an optional parameter named use_gpu in predict interface which is
set to TRUE by default. If use_gpu is true, then GPU support on predict
is enabled.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -82,6 +82,8 @@ def _strip_trailing_nulls_from_class_values(class_values):
 
 def predict(schema_madlib, model_table, test_table, id_col,
             independent_varname, output_table, pred_type, use_gpu, **kwargs):
+    if not pred_type:
+        pred_type = 'response'
     input_validator = PredictInputValidator(
         test_table, model_table, id_col, independent_varname,
         output_table, pred_type, MODULE_NAME)
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -30,7 +30,7 @@ import numpy as np
 from madlib_keras_helper import expand_input_dims
 from madlib_keras_helper import MODEL_DATA_COLNAME
 from madlib_keras_validator import PredictInputValidator
-from madlib_keras_wrapper import compile_and_set_weights
+from madlib_keras_wrapper import set_model_weights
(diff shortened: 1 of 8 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/851/

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -112,6 +114,9 @@ def predict(schema_madlib, model_table, test_table, id_col,
         class_values, intermediate_col, pred_col_name,
         pred_col_type, is_response, MODULE_NAME)
 
+    segment_id = -1 if is_platform_pg() else '{0}.gp_segment_id'.format(
```
kaknikhil: why do we need to qualify the segment_id with the table name?

njayaram2: The query that this gets used in selects from both `{test_table}` and `{model_table}`. `gp_segment_id` is an implicit column in both these tables that would lead to ambiguous column name.

kaknikhil: Makes sense. Maybe we should create a function named `get_segment_id` which will replace all instances of this if check ? It will make our code less error prone to the situation you described.

njayaram2: Sure, but the only other instance is in fit, and there we don't have to prefix it with a table name.
Btw, the value of this variable for postgres makes no difference (it could be -1 or any other integer), since we do not infer it is postgres using this value anywhere else in the code. It just shouldn't be `gp_segment_id` for postgres.

asfgit:
https://builds.apache.org/job/madlib-pr-build/859/

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
njayaram2: Realized that if we use the refactored function, we will have to call it
twice in fit function, once for train table and once for validation
table, since the function prefixes the table name for GPDB. I think it's
fine the way it was now, we can refactor it later if the use for this check
gets more ubiquitous. So reverted the refactoring commit, and will keep
it this way for now.

asfgit:
https://builds.apache.org/job/madlib-pr-build/862/

PR #379, opened 2019-04-30
 DL: Replace use_gpu flag with gpus_per_host integer 

kaknikhil:
JIRA: MADLIB-1308

Previously, gpus_per_host were hard coded to 4. This commit removes this
hard coding and takes in this value from the user.

We also tried to use the tensorflow function `list_local_devices` to get
the count of gpus per host. This did give us the count but would hang
forever on some segments. So we decided to not use this function.

We now cache the CUDA_VISIBLE_DEVICES env variable (which is set to
-1 for master) and then reset it at the end of fit function.

Also fixed a bug  in get device name function for postgres

PR #383, opened 2019-05-02
Dl/remove hardcode

kaknikhil:
Previous PR for reference https://github.com/apache/madlib/pull/379

JIRA: MADLIB-1308

Previously, gpus_per_host were hard coded to 4. This commit removes this
hard coding and takes in this value from the user.

We also tried to use the tensorflow function list_local_devices to get
the count of gpus per host. This did give us the count but would hang
forever on some segments. So we decided to not use this function.

We now cache the CUDA_VISIBLE_DEVICES env variable (which is set to
-1 for master) and then reset it at the end of fit function.

Finally, we dynamically calculate the gpu memory fraction to support
the case when the number of gpus is less than the number of segments.

Also fixed a bug in get device name function for postgres

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
@@ -38,7 +38,7 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.madlib_keras_fit(
     compile_params          VARCHAR,
     fit_params              VARCHAR,
     num_iterations          INTEGER,
-    use_gpu                 BOOLEAN,
+    gpus_per_host           INTEGER,
     validation_table        VARCHAR,
     name                    VARCHAR,
     description             VARCHAR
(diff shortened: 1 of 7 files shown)


kaknikhil:
@reductionista The code changes are ready to be re-reviewed. 

For this comment https://github.com/apache/madlib/pull/379#discussion_r279906423, we removed the duplicated set code but kept the two functions to make the code easier to reflect that we are setting the value at the begining and resetting it later.

asfgit:
https://builds.apache.org/job/madlib-pr-build/872/

src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
```diff
@@ -158,12 +161,13 @@ CREATE AGGREGATE MADLIB_SCHEMA.fit_step(
...
     /* compile_params */         TEXT,
     /* fit_params */             TEXT,
-    /* use_gpu */                BOOLEAN,
+    /* gpus_per_host  */         INTEGER,
+    /* segments_per_host  */     INTEGER,
     /* previous_state */         BYTEA
 )(
     STYPE=BYTEA,
     SFUNC=MADLIB_SCHEMA.fit_transition,
-    m4_ifdef(`__POSTGRESQL__', `', `prefunc=MADLIB_SCHEMA.fit_merge,')
```
reductionista: I'm not sure if this was here before and I missed it, or if it's new.  Looks like this would break postgres support, but I haven't tested it--was this change intentional?

reductionista:
As long as this compiles and runs on postgres platform, LGTM

src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
kaknikhil: Turns out the postgres throws a warning saying `aggregate attribute "prefunc" not recognized` but does not fail. I think it's safer to have the if check so i will revert this to the original m4 ifdef.

asfgit:
https://builds.apache.org/job/madlib-pr-build/881/

asfgit:
https://builds.apache.org/job/madlib-pr-build/882/

asfgit:
https://builds.apache.org/job/madlib-pr-build/883/

PR #391, opened 2019-05-15
DL: Refactor code to move some constants to helper file

njayaram2:
Define MINIBATCH_OUTPUT_DEPENDENT_COLNAME_DL and
MINIBATCH_OUTPUT_INDEPENDENT_COLNAME_DL in madlib_keras_helper.py_in
instead of in input_data_preprocessor.py_in. Importing these from
input_data_preprocessor was also doing other imports that are not
necessary for madlib_keras.py_in.


asfgit:
https://builds.apache.org/job/madlib-pr-build/912/

PR #394, opened 2019-05-17
DL: Add model_arch column to model data table

khannaekta:
JIRA: MADLIB-1347

This PR adds the 'model_arch' column to the model data table so that
predict and evaluate can directly get the model_arch from the model data
table instead of the model arch table.

We have made changes to the predict code in this commit but the changes
for evaluate will be done in a future PR.

Co-authored-by: Nikhil Kak <nkak@pivotal.io>

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
@@ -320,9 +320,10 @@ def fit(schema_madlib, source_table, model,model_arch_table,
                   description, aggregate_runtime, class_values])
 
     create_output_table = plpy.prepare("""
-        CREATE TABLE {0} AS
-        SELECT $1 as model_data""".format(model), ["bytea"])
-    plpy.execute(create_output_table, [model_state])
+        CREATE TABLE {0} AS SELECT
+        $1 as model_data,
+        $2 as {1}""".format(model, Format.MODEL_ARCH), ["bytea", "json"])
+    plpy.execute(create_output_table, [model_state, model_arch])
 
     if is_platform_pg():
         clear_keras_session()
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/916/

src/ports/postgres/modules/deep_learning/predict_input_params.py_in
```diff
@@ -36,28 +34,11 @@ class PredictParamsProcessor:
...
-            WHERE {2} = {3}
-        """.format(Format.MODEL_ARCH, model_arch_table, Format.MODEL_ID,
-                   model_arch_id)
-        query_result = plpy.execute(model_arch_query)
-        if not query_result or len(query_result) == 0:
-            plpy.error("{0}: No model arch found in table {1} with id {2}".format(
-                self.module_name, model_arch_table, model_arch_id))
-        return query_result[0]
+    def _get_dict_for_table(self, table_name):
+        return plpy.execute("SELECT * FROM {0}".format(table_name))[0]
```
reductionista: Since we don't ever validate the number of rows in these tables, maybe we should add a LIMIT 1 just in case?

src/ports/postgres/modules/deep_learning/madlib_keras.py_in
```diff
@@ -321,9 +321,10 @@ def fit(schema_madlib, source_table, model,model_arch_table,
                   description, aggregate_runtime, class_values])
 
     create_output_table = plpy.prepare("""
-        CREATE TABLE {0} AS
-        SELECT $1 as model_data""".format(model), ["bytea"])
-    plpy.execute(create_output_table, [model_state])
+        CREATE TABLE {0} AS SELECT
+        $1 as model_data,
+        $2 as {1}""".format(model, Format.MODEL_ARCH), ["bytea", "text"])
```
reductionista: I think the type should be JSON rather than TEXT:

$2::JSON as {1} ...

kaknikhil: you are right it should be json. I will fix it

src/ports/postgres/modules/deep_learning/predict_input_params.py_in
kaknikhil: we thought about it as well but i guess it doesn't hurt add limit 1

asfgit:
https://builds.apache.org/job/madlib-pr-build/920/

PR #401, opened 2019-05-30
DL: fix default_buffer_size calculation

jingyimei:
JIRA: MADLIB-1340

In buffer size calculator, we assumed the array to be packed is one dimention, and use the array_upper(x, 1) to get the length
of the array. In DL input data preprocessor, we passed the first element
in array_ndims as the length, which is not right because the
array can be multi-dimentional. This caused
default_buffer_size_calculator returns a bigger buffer size than it
supposed to have and rows get too big and crashes database.

Instead, we should use
product of all the elements from array_ndims to represent the actual length of
the array. For example, if array_ndims returns [32,32,3], we should pass
32*32*3 instead of 32. This commit fixes this issue by passing the
right length to defualt buffer calculator.

### DIFF
--- a/src/ports/postgres/modules/internal/db_utils.py_in
+++ b/src/ports/postgres/modules/internal/db_utils.py_in
@@ -26,8 +26,7 @@ m4_changequote(`<!', `!>')
 QUOTE_DELIMITER="$__madlib__$"
 
 
-def get_distinct_col_levels(source_table, col_name, col_type=None,
-                            include_nulls=False):
+def get_distinct_col_levels(source_table, col_name, col_type=None, include_nulls=False):
     """
     Add description here
     :return:
@@ -93,3 +92,28 @@ def is_col_1d_array(source_table, col_name):
     """.format(col_name, source_table)
     result = plpy.execute(query)
     return result[0]["n_y"]
+
+# ------------------------------------------------------------------------------
+# This function runs postgres array_ndims function to get
+# the dimension of an array. For example if it is a 3
+# dimension array it will be an array with 3 elements
+# like [32,32,3].
+def get_ndims(source_table, col_name):
+    array_ndims = plpy.execute("SELECT array_ndims({0}) AS ndims FROM {1}".
+        format(col_name, source_table))[0]['ndims']
+
+    return array_ndims
+
+# This function is to calculate the total `length` of a
+# multi dimentional array. For example, if an array is
+# with 3 dimentions and ndims=[32,32,3], this function
+# will return the product of them, which is 32*32*3
+def get_product_of_dimensions(source_table, col_name):
+    ndims = get_ndims(source_table, col_name)
+    dimension_product = 1
+    for i in range(1, ndims + 1):
+        array_upper_query = "SELECT array_upper({0}, {1}) AS dimension from {2}".format(col_name, i, source_table)
+        dimension = plpy.execute(array_upper_query, 1)[0]['dimension']
+        dimension_product *= dimension
+
+    return dimension_product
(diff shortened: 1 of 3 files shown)


fmcquillan99:
That's a good catch.

asfgit:
https://builds.apache.org/job/madlib-pr-build/947/

src/ports/postgres/modules/internal/db_utils.py_in
```diff
@@ -93,3 +92,28 @@ def is_col_1d_array(source_table, col_name):
...
+# like [32,32,3].
+def get_ndims(source_table, col_name):
+    array_ndims = plpy.execute("SELECT array_ndims({0}) AS ndims FROM {1}".
+        format(col_name, source_table))[0]['ndims']
+
+    return array_ndims
+
+# This function is to calculate the total `length` of a
+# multi dimentional array. For example, if an array is
+# with 3 dimentions and ndims=[32,32,3], this function
```
khannaekta: `dimentions` -> `dimensions`

src/ports/postgres/modules/internal/db_utils.py_in
```diff
@@ -93,3 +92,28 @@ def is_col_1d_array(source_table, col_name):
...
+# will return the product of them, which is 32*32*3
+def get_product_of_dimensions(source_table, col_name):
+    ndims = get_ndims(source_table, col_name)
+    dimension_product = 1
+    for i in range(1, ndims + 1):
+        array_upper_query = "SELECT array_upper({0}, {1}) AS dimension from {2}".format(col_name, i, source_table)
```
khannaekta: We can also call `get_col_dimension(col_name, source_table, i)` which calls the same query internally.
But, will leave it upto you whichever way you prefer.

asfgit:
https://builds.apache.org/job/madlib-pr-build/948/

reductionista:
I just want to comment that we should make sure this is tested for 16-bit and 8-bit data types.  If the logic is just based on the size being < 1GB then it will crash whenever there are more than 134217728 elements in the batched array.  (134217728 is the hardcoded maximum number of elements that can be in any array for all of postgres/greenplum).  For 32-bit data types, that ends up being the same as the < 1GB requirement, but for 16-bit it's < 512MB, and for 8-bit it's < 256MB. 

I don't think we can close  this JIRA (https://issues.apache.org/jira/browse/MADLIB-1340) until we're sure both of these issues are handled properly.

jingyimei:
> I just want to comment that we should make sure this is tested for 16-bit and 8-bit data types. If the logic is just based on the size being < 1GB then it will crash whenever there are more than 134217728 elements in the batched array. (134217728 is the hardcoded maximum number of elements that can be in any array for all of postgres/greenplum). For 32-bit data types, that ends up being the same as the < 1GB requirement, but for 16-bit it's < 512MB, and for 8-bit it's < 256MB.
> 
> I don't think we can close this JIRA (https://issues.apache.org/jira/browse/MADLIB-1340) until we're sure both of these issues are handled properly, as there may be two different bugs here.

@reductionista That's good point. Let's merge this pr first, since it still improves robustness, and raise another PR to address different bits maximum number of elements issue later.

src/ports/postgres/modules/internal/db_utils.py_in
```diff
@@ -93,3 +92,28 @@ def is_col_1d_array(source_table, col_name):
...
     result = plpy.execute(query)
     return result[0]["n_y"]
+
+# ------------------------------------------------------------------------------
+# This function runs postgres array_ndims function to get
+# the dimension of an array. For example if it is a 3
+# dimension array it will be an array with 3 elements
+# like [32,32,3].
+def get_ndims(source_table, col_name):
+    array_ndims = plpy.execute("SELECT array_ndims({0}) AS ndims FROM {1}".
```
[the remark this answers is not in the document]
njayaram2: Yes, that's right. Will do `plpy.execute(query_str, 1)`.

asfgit:
https://builds.apache.org/job/madlib-pr-build/950/

PR #404, opened 2019-06-04
Fix tensorflow crashes

reductionista:
This adds the `std::_Rb_tree<std::string>` functions defined in madlib
to the list of exported symbols in `src/unexported_symbols_list.txt` (for darwin) and `src/library.ver` (for all other platforms).

This fixes an intermittent crash that was happening in the deep_learning module
in different forms on different platforms, due to tensorflow calling madlib's
instantiation of these functions, which circumvents the hiding of madlib's
global new and delete symbols (intended to override the libstdc++ version
of these symbols for all of `libmadlib.so`, but not for anything outside of it).

These `_Rb_tree` functions are included in madlib due to use of `std::set<string>`
in `madlib/src/modules/linalg/metric.cpp`.  Tensorflow is a python
library, but it loads Google's C++ protobuf library `_message.so`.  The
crash was happening because `_message.so` also uses `std::set<string>` for
some things.


asfgit:
https://builds.apache.org/job/madlib-pr-build/958/

asfgit:
https://builds.apache.org/job/madlib-pr-build/959/

PR #407, opened 2019-06-07
DL: Create one keras session per iteration for pg in fit and evaluate 

kaknikhil:
JIRA: MADLIB-1356

Currently for postgres, we create one keras session in
fit/evaluate/predict and keep it open for all the iterations and then
close it at the end.  We did a few experiments and found out that if we
create 1 keras session per iteration , it performs slightly better.
So this commit modifies the code to make pg behave the same as gpdb(1
session per iteration) in fit_transition, eval_transition and
internal_predict functions.

We also changed set_keras_session to use a device_name to be consistent
with other keras operations.

This commit also fixes the following bug: 
If gpus are available on the
host but the user passed in 0 for the gpus_per_host param, we still end
up using gpu memory. This is because in the madlib_keras.fit UDF we call
model.get_weights() to initialize the model and this function ends up
using gpu(s) if there are any available.

Also refactored some image count related code

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
@@ -110,20 +110,10 @@ def fit(schema_madlib, source_table, model, model_arch_table,
     num_classes = get_num_classes(model_arch)
     fit_validator.validate_input_shapes(input_shape)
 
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/965/

asfgit:
https://builds.apache.org/job/madlib-pr-build/972/

PR #413, opened 2019-06-14
Pivot: Fix array_agg + distinct scaling issue on gpdb

orhankislal:
JIRA: MADLIB-1361

With large datasets, pivot fails because of the array_agg(distinct)
query. This is because array_agg collects the values first and filters
the distinct values later. This causes the array_agg to go out of
memory.

This commit fixes the issue by separating distinct from array_agg.
We use a subquery to get the distinct values. Then we aggregate these
values using array_agg.

Closes #413

Co-authored-by: Ekta Khanna <ekhanna@pivotal.io>


asfgit:
https://builds.apache.org/job/madlib-pr-build/979/

PR #420, opened 2019-07-10
Fix Design Doc

reductionista:
The `\algorithm` macro can optionally have input params specified.  But for some reason, they become mandatory if you want to `\label` the algorithm.

Leaving off the input params from a labelled algorithm block results in pdflatex thinking there is an extra unmatched `}` found, and failing to recover from the confusion.  (At least in the version of pdflatex I have, not sure if this happens in all LaTex environments.)

Changes:

- Add input params to last part of WCC impl.

- Add `alg:` prefix to label, for consistency with rest of doc

- Change name of algorithm from "Update oldupdate table" to
"Update message table" to match rest of text.

- Minor typo correction (uppercase).

### DIFF
--- a/doc/design/modules/graph.tex
+++ b/doc/design/modules/graph.tex
@@ -125,7 +125,7 @@ negative cycle in the graph.
 
 
 \begin{algorithm}[SSSP$(V,E,start)$] \label{alg:sssp}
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/990/

PR #422, opened 2019-07-11
Correlation: Process deconstruction in chunks for grouping

njayaram2:
JIRA: MADLIB-1301
While deconstructing the correlation matrix to create the output table,
a big UNION ALL query was created, with one sub-query for each distinct
grouping value. This was causing the memory, stack, and performance
related issues.
The fix is to run multiple queries, with each query processing the
deconstruction of the correlation matrix for a limited number of
groups (we have defaulted the value to 10). This value can be
parameterized by the user with a newly introduced optional parameter
named `n_groups_per_run` for both correlation and covariance.

Co-authored-by: Orhan Kislal <okislal@apache.org>

### DIFF
--- a/src/ports/postgres/modules/stats/correlation.py_in
+++ b/src/ports/postgres/modules/stats/correlation.py_in
@@ -363,7 +363,7 @@ def _create_deconstruction_query(schema_madlib, col_names, grouping_cols,
 
     Returns:
         List of Strings where each string is a SQL sub-query for deconstructing
-        the matrix. Each sub-query covers `n_groups_per_run` number of union all
+        the matrix. Each sub-query covers n_groups_per_run number of union all
         queries
     """
     # The matrix that holds the PCC computation must be converted to a
--- a/src/ports/postgres/modules/stats/correlation.py_in
+++ b/src/ports/postgres/modules/stats/correlation.py_in
@@ -9,6 +9,7 @@ from time import time
 
 import plpy
 from utilities.control import MinWarning
+from utilities.utilities import _assert
 from utilities.utilities import add_postfix
 from utilities.utilities import get_table_qualified_col_str
 from utilities.utilities import py_list_to_sql_string
@@ -20,10 +21,9 @@ from utilities.validate_args import get_cols_and_types
 from utilities.validate_args import input_tbl_valid
 from utilities.validate_args import output_tbl_valid
 
-
 def correlation(schema_madlib, source_table, output_table,
                 target_cols, grouping_cols, get_cov=False,
-                verbose=False, **kwargs):
+                verbose=False, n_groups_per_run=10, **kwargs):
     """
     Populates an output table with the coefficients of correlation between
     the columns in a source table
@@ -45,10 +45,12 @@ def correlation(schema_madlib, source_table, output_table,
         _numeric_column_names, _nonnumeric_column_names = _get_numeric_columns(source_table)
         _target_cols = _analyze_target_cols(source_table, target_cols,
                                             function_name)
+        if n_groups_per_run is None:
+             n_groups_per_run = 10
         # Validate grouping_cols param
         if grouping_cols:
             _validate_grouping_cols_param(source_table, grouping_cols,
-                                          function_name)
+                                          function_name, n_groups_per_run)
         if _target_cols:
             # prune all non-numeric column types from target columns
             _existing_target_cols = []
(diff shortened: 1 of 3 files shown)


fmcquillan99:
<dt>n_groups_per_run (optional)</dt>
<dd>INTEGER, default: 10. Number of groups to deconstruct at a time.
This parameter value is ignored if grouping_cols is not specified.
Processing large number of groups together may have memory/stack/execution time
issues. This parameter should be used with caution. The number of groups that
can be processed together is a function of various database settings.</dd>
</dl>

This is hard to understand.  Can we give guidance on when to 
change it from the default?  Need to make the trade off clear.

E.g., 
increasing does x at the expense of y
decreasing does a at the expense of b

Also I don't know what to call it yet since I'm not 100% sure what it does.

asfgit:
https://builds.apache.org/job/madlib-pr-build/991/

src/ports/postgres/modules/stats/test/correlation.sql_in
```diff
@@ -134,3 +147,16 @@ SELECT assert(relative_error(temperature, 17.1500) < 0.0001, 'calculated covaria
 SELECT assert(relative_error(humidity, -28.7000) < 0.0001, 'calculated covariance value ' || humidity || 'does not match expected: ' || -28.7000) FROM example_data_gr2_output_covariance WHERE variable ='new_col' AND gr='1' AND gr2='1';
 
 SELECT assert(relative_error(id, 15.4688) < 0.0001, 'calculated covariance value ' || id || 'does not match expected: ' || 15.4688) FROM example_data_gr2_output_covariance WHERE variable ='new_col' AND gr='1' AND gr2='1';
+
+DROP TABLE IF EXISTS example_data_gr2_output_covariance, example_data_gr2_output_covariance_summary;
+SELECT covariance( 'example_data_gr2',
+                    'example_data_gr2_output_covariance',
+                    'temperature, humidity, id, new_col',
+                    FALSE,
+                    'gr,gr2', 50);
```
makemebitter: Do you think it makes sense to assert the count of the output table when grouping is used ?

src/ports/postgres/modules/stats/correlation.py_in
```diff
@@ -330,9 +355,11 @@ def _create_deconstruction_query(schema_madlib, col_names, grouping_cols,
                                     the matrix to deconstruct
         @param cor_mat              Name of column that containss the matrix
                                     to deconstruct
+        @param n_groups_per_run     Number of groups to deconstruct in a single
+                                    sub-query
 
     Returns:
-        String (SQL querry for deconstructing the matrix)
+        List of Strings (each string is a SQL sub-query for deconstructing the matrix)
```
makemebitter: maybe also add that each subquery in the return list contains n groups.

src/ports/postgres/modules/stats/correlation.py_in
```diff
@@ -291,8 +296,28 @@ def _populate_output_table(schema_madlib, source_table, output_table,
...
             USING (column_position)
-            """.format(num_cols=len(col_names), **locals())
+        """
+
+        # Create the output table.
+        create_output_table_query = """
+            CREATE TABLE {output_table} AS
+            {select_deconstruct_query}
+            """.format(**locals()).format(
+                deconstruction_query=deconstruction_query_list[0],
```
makemebitter: consider adding a comment to explain why we create the table using the first subquery. I noticed that there is a comment on line 309 but we should make it more clear and move it before the create table query.

orhankislal:
@fmcquillan99 
Increasing this value means we construct a larger `union all` query which uses more memory. We also noticed that large union all queries slow down the execution.
Decreasing this value means we issue a large number of `plpy.execute` commands. After a certain point, the overhead starts to be more significant.
If you have a large number of groups with small chunks of data, it might be helpful to slightly increase this number (and vice versa). However, this is not a definitive rule. We introduced this change because having a limited number of `union all` queries (as opposed to to processing all of the groups at once) was helpful for large number of groups. It is sensitive to the groups as well as the system parameters and the user should experiment with it.

src/ports/postgres/modules/stats/correlation.py_in
```diff
@@ -291,8 +296,28 @@ def _populate_output_table(schema_madlib, source_table, output_table,
                 {deconstruction_query}
             ) {matrix_subquery}
             USING (column_position)
-            """.format(num_cols=len(col_names), **locals())
+        """
```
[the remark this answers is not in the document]
orhankislal: Because this string is also used in Line 315. There, we format with `deconstruction_query=deconstruction_query_list[i]`

asfgit:
https://builds.apache.org/job/madlib-pr-build/992/

asfgit:
https://builds.apache.org/job/madlib-pr-build/993/

fmcquillan99:
@orhankislal thanks for the explanation

LGTM

asfgit:
https://builds.apache.org/job/madlib-pr-build/995/

PR #423, opened 2019-07-16
Association Rules: Improve performance

njayaram2:
JIRA: MADLIB-1327
Assocation rules was slow due to a blow up in number of candidate
itemsets before checking for their support and graduating them to
frequent itemsets. This PR changes ensures the candidate itemsets are
ordered, and only a few consecutive itemsets must be considered to merge
and create a new candidate itemset for checking support. This also
reduces the result of a join query significantly (previously, the result
of the join were all potential candidates, which were redundant and
which may have resulted in itemsets that were larger than what was
considered in the current iteration).
This commit also adds relevant tests in dev-check.

Co-authored-by: Orhan Kislal <okislal@apache.org>

### DIFF
--- a/src/modules/assoc_rules/assoc_rules.cpp
+++ b/src/modules/assoc_rules/assoc_rules.cpp
@@ -23,6 +23,8 @@ typedef struct perm_fctx
     char*    positions;
     int32    pos_len;
     int32    num_elems;
+    int32    max_LHS_size;
+    int32    max_RHS_size;
     int32    num_calls;
 
     /* type information for the result type*/
@@ -38,6 +40,8 @@ typedef struct perm_fctx
  * @param args      Two-element array.
  *                  args[0] is the text form of a closed frequent pattern.
  *                  args[1] is the number of items in the pattern.
+                    args[2] is the max number of elements in the lhs of the rule
+                    args[3] is the max number of elements in the rhs of the rule
  * @param max_call  The number of  will be generated.
  *
  * @return  The struct including the variables which will be used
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/madlib-pr-build/994/

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -109,6 +117,29 @@ gen_rules_from_cfp::SRF_next(void *user_fctx, bool *is_last_call) {
...
+        // flags[i]=True means that element is on the lhs (and vice versa)
+        for (i = 0; i < myfctx->num_elems; ++i) {
+            if (!myfctx->flags[i]) {
+                countRHS ++;
+            } else {
+                countLHS ++;
+            }
+        }
+        if (countLHS > myfctx->max_LHS_size || countRHS > myfctx->max_RHS_size){
+            --myfctx->num_calls;
```
kaknikhil: Can you explain this in a comment ?

src/ports/postgres/modules/assoc_rules/test/assoc_rules.sql_in
```diff
@@ -137,29 +144,63 @@ begin
           assoc_array_eq(t1.post, t2.post) AND
           abs(t1.support - t2.support) < 1E-10 AND
           abs(t1.confidence - t2.confidence) < 1E-10;
+    IF (result2 = 'FAIL') THEN
```
kaknikhil: Do we have tests that check for the support and confidence values ? If not, it might be a good time to add them to prevent any future regressions

orhankislal: I am not sure what you mean. This is literally the test that check for the support and confidence for each rule. It counts the rules that match the fixed values in the expected results table `test2_exp_result` and writes PASS or FAIL to a table if this count is 7. Unless the rules as well as the support and confidence levels match, the result will be a failure. I guess it is a bit unconventional among madlib tests and we might think about changing it in the future.

asfgit:
https://builds.apache.org/job/madlib-pr-build/996/

src/ports/postgres/modules/assoc_rules/test/assoc_rules.sql_in
kaknikhil: ohh I see I just saw the exception and thought this was testing some error condition.

fmcquillan99:
LGTM

see JIRA https://issues.apache.org/jira/browse/MADLIB-1327 for scale testing, which shows significant improvement when using the new params

asfgit:
https://builds.apache.org/job/madlib-pr-build/997/

PR #435, opened 2019-09-04
Separate mst module from fit_multiple_models module

reductionista:
Very simple PR:

Rename `madlib_keras_fit_multiple_model.*` to `madlib_keras_model_selection.*` so that once we merge in the `fit_multiple_models` code from #425, they will each have separate user-doc pages.  This will allow more flexibility to expand the current helper function into a more advanced model management utility.

Other minor changes:

- improves validation of input params `fit_params_list` and `compile_params_list` (checks for NULL values, reports error).
- added `.vscode` to `.gitignore` for anyone using Visual Studio Code


src/ports/postgres/modules/deep_learning/madlib_keras_validator.py_in
```diff
@@ -352,6 +352,10 @@ class MstLoaderInputValidator():
             )
         )
     def _validate_compile_and_fit_params(self):
+        if not self.fit_params_list:
+            plpy.error(
```
kaknikhil: why break up this line of code in three, why not just do
``` 
plpy.error("fit_params_list cannot be NULL")
```

src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.sql_in
```diff
@@ -18,7 +18,7 @@
  * under the License.
  *
  *
- * @file madlib_keras_fit_multiple_model.sql_in
+ * @file madlib_keras_model.sql_in
```
kaknikhil: Shouldn't this be madlib_keras_model_selection?

kaknikhil:
LGTM. Left a couple of minor comments

PR #430, opened 2019-08-07
Add generate mst table utility.

makemebitter:
JIRA: MADLIB-1375
This commit adds a new function to generate the model selection table.
It generates such tables as the naive combinations of the three input
lists of parameter choices. This utility also validates and
de-duplicates the inputs.

Co-authored-by: Domino Valdano domino@apache.org

PR #437, opened 2019-09-05
DL: Remove model_arch_table from output of load_model_selection_table()

makemebitter:
JIRA: MADLIB-1381

The utility function `load_model_selection_table()` assumes that there
is only 1 model architecture table (passed in as a parameter).
Therefore, it should not be added to the output table.
This PR removes `model_arch_table` column from the output table of
`load_model_selection_table()` and creates a summary table with this
column instead.
Also, added tests and split the unit test file.


fmcquillan99:
I re-ran some tests from https://github.com/apache/madlib/pull/430 and seems OK.

LGTM

PR #439, opened 2019-09-06
DL: Add support in preprocessor to evenly distribute data for GPDB

khannaekta:
JIRA: MADLIB-1378

Currently, `training_preprocessor_dl` and `validation_preprocessor_dl`
doesn't guarantee even distribution of the data among segments,
especially when the number of buffers is not much larger than the number
of segments. This commit adds support to the preprocessor so that it
always distributes the data as evenly as possible among the segments.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
+++ b/src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
@@ -18,6 +18,7 @@
  * under the License.
  *
  *//* ----------------------------------------------------------------------- */
+m4_include(`SQLCommon.m4')
 
 DROP TABLE IF EXISTS data_preprocessor_input;
 CREATE TABLE data_preprocessor_input(id serial, x double precision[], label TEXT);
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
```diff
@@ -88,6 +89,23 @@ SELECT training_preprocessor_dl(
...
+m4_ifdef(<!__POSTGRESQL__!>, <!!>, <!
+DROP TABLE IF EXISTS data_preprocessor_input_batch, data_preprocessor_input_batch_summary;
+SELECT training_preprocessor_dl(
+  'data_preprocessor_input',
+  'data_preprocessor_input_batch',
+  'id',
+  'x',
+  1);
+
+SELECT assert(count(*)=(SELECT ceil(17.0/count(*)) from gp_segment_configuration WHERE role = 'p' and content != -1), 'Even distribution of buffers failed.')
```
orhankislal: What is the significance of 17? We should explain it in a comment.

src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
```diff
@@ -88,6 +89,23 @@ SELECT training_preprocessor_dl(
...
+FROM data_preprocessor_input_batch
+WHERE gp_segment_id = 0
+GROUP BY gp_segment_id;
```
orhankislal: If we have the `WHERE gp_segment_id = 0` clause, why do we need the `group by`?

src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
khannaekta: It is for the total number of images in the source table.
Will add an comment explaining it. Thanks.

src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
khannaekta: Good catch! Initially we were planning to count images for all the segments, but now we only want to assert on a single segment so `GROUP BY ` doesn't make sense. Will update the test query to remove it.

PR #444, opened 2019-09-23
DL: Improve performance for madlib_keras_predict()

khannaekta:
JIRA: MADLIB-1345

Passing huge model weights as a param to `internal_keras_predict()` for
each row in the table slows the performance of overall `madlib_keras_predict()`.
With this PR, the model weights will be passed in as a param only for the
very first row(min(ctid)) fetched on each segment and NULL for the rest.
With this change, we see `~4x` performance boost in the execution time
of the existing `madlib_kerase_predict()`.


khannaekta:
Jenkins: retest this please

orhankislal:
retest this please

PR #445, opened 2019-09-30
updated DL preprocessor docs for bytea

fmcquillan99:
Since we changed to bytea for packed dep and indep vars, needed to update the users docs to correspond.

For weird and unknown doxygen issues, @kaknikhil & I learned that 
```
DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.agg_array_concat(anyarray);
CREATE AGGREGATE MADLIB_SCHEMA.agg_array_concat(anyarray) (
   SFUNC = array_cat,
   PREFUNC = array_cat,
   STYPE = anyarray
   );
```
works but 
```
DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.agg_array_concat(anyarray);
CREATE AGGREGATE MADLIB_SCHEMA.agg_array_concat(anyarray) (
   SFUNC = array_cat,
   STYPE = anyarray,
   PREFUNC = array_cat
   );
```
results in a blank page being published.  This is not root cause, but the docs do compile by doing this.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
+++ b/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
@@ -216,6 +216,10 @@ validation_preprocessor_dl(source_table,
       <tr>
         <th>independent_var</th>
         <td>BYTEA. Packed array of independent variables in PostgreSQL bytea format.
+        Arrays of independent variables packed into the output table are
+        normalized by dividing each element in the independent variable array by the
+        optional 'normalizing_const' parameter.  Training data is shuffled, but
+        validation data is not.
         </td>
       </tr>
       <tr>
--- a/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
+++ b/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
@@ -18,7 +18,7 @@
  * under the License.
  *
  * @file input_preprocessor_dl.sql_in
- * @brief TODO
+ * @brief Utilities to prepare input image data for use by deep learning modules.
  * @date December 2018
  *
  */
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
```diff
@@ -209,25 +214,39 @@ validation_preprocessor_dl(source_table,
     validation_preprocessor_dl() contain the following columns:
     <table class="output">
       <tr>
-        <th>buffer_id</th>
-        <td>INTEGER. Unique id for each row in the packed table.
+        <th>independent_var</th>
+        <td>BYTEA. Packed array of independent variables in PostgreSQL bytea format.
```
kaknikhil: We can also add that the independent var is first normalized and then packed and converted to bytea. I understand that we already mentioned this in the function definition. But I believe that it will be valuable to mention it again just like we do for dependent_var

src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
```diff
@@ -209,25 +214,39 @@ validation_preprocessor_dl(source_table,
...
+        <td>INTEGER[]. Shape of the independent variable array after preprocessing.
+        The first element is the number of images packed per row, and subsequent
+        elements will depend on how the image is described (e.g., channels first or last).
+        </td>
+      </tr>
+      <tr>
+        <th>dependent_var_shape</th>
+        <td>INTEGER[]. Shape of the dependent variable array after preprocessing.
+        The first element is the number of images packed per row, and the second
+        element is the number of class values.
```
kaknikhil: Do we need to mention anything about one hot encoding here ?

src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
fmcquillan99: added this comment

src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
fmcquillan99: We do talk about 1-hot encoding in some detail when talking about `dependent_varname` in the function definition, so I am hoping that is sufficient.  If not, please let me know.

PR #448, opened 2019-10-04
SVM: Lower bound the default for n_components

orhankislal:


fmcquillan99:
Does this new default get applied to:
```
svm_classification()
svm_regression()
svm_one_class()
```
?

orhankislal:
> Does this new default get applied to:
> 
> ```
> svm_classification()
> svm_regression()
> svm_one_class()
> ```
> 
> ?

Yes, they all call `_extract_kernel_params` at some point to read the variable which sets the default.

fmcquillan99:
LGTM

PR #449, opened 2019-10-09
DL: Use REAL[] instead of anyarray for aggregating arrays

kaknikhil:
JIRA: MADLIB-1334

Using real[] instead of anyarray for agg_array_concat solves the scaling
problem i.e. running input preprocessor on same dataset with different
buffer sizes results in comparable runtimes. This is because the plan
for the agg with real[] aggregates on the segments and then gathers on
the master but the plan for the agg with anyarray aggregate first
gathers all the data on the master and then runs the aggregate function
on the master.

Here are the results from my local mac with gpdb 5.21 3 segments

```
mnist 10k rows anyarray agg
madlib=# select madlib.training_preprocessor_dl('mnist_train_10k', 'mnist_batch_10k_anyarray', 'y', 'x');
Time: 109790.096 ms

mnist 10k rows real_array_agg
madlib=# select madlib.training_preprocessor_dl('mnist_train_10k', 'mnist_batch_10k_real_array', 'y', 'x');
Time: 14086.535 ms

mnist 60k rows with anyarray agg
madlib=# select madlib.training_preprocessor_dl('mnist_train', 'mnist_batch_anyarray', 'y', 'x');
Had to cancel after 55 mins

mnist 60k rows with real_array agg
madlib=# select madlib.training_preprocessor_dl('mnist_train', 'mnist_batch_real_array', 'y', 'x');
Time: 580822.559 ms
```

Co-authored-by: Domino Valdano <dvaldano@pivotal.io>

### DIFF
--- a/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
+++ b/src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
@@ -814,12 +814,11 @@ $$ LANGUAGE plpythonu VOLATILE
 m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `MODIFIES SQL DATA', `');
 
 
-DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.agg_array_concat(anyarray);
-CREATE AGGREGATE MADLIB_SCHEMA.agg_array_concat(anyarray) (
+DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.agg_array_concat(REAL[]);
+CREATE AGGREGATE MADLIB_SCHEMA.agg_array_concat(REAL[]) (
    SFUNC = array_cat,
    PREFUNC = array_cat,
-   STYPE = anyarray
-
+   STYPE = REAL[]
    );
 
 CREATE FUNCTION MADLIB_SCHEMA.convert_array_to_bytea(var REAL[])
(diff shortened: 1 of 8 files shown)


kaknikhil:
@reductionista I agree that dependent_var should stay as smallint. I have made changes to the code to take care of this, can you re-review ?

kaknikhil:
Retest this please

src/ports/postgres/modules/deep_learning/input_data_preprocessor.sql_in
```diff
@@ -849,3 +857,13 @@ import numpy as np
 return np.frombuffer(var, dtype=np.float32)
```
reductionista: I must have missed this last time, but shouldn't this be `tobytes()` instead of `frombuffer()`?  (For function `convert_array_to_bytea()`)

Also, maybe we should rename it to `convert_real_array_to_bytea()` now that we have two different array types.

src/ports/postgres/modules/deep_learning/test/madlib_keras_cifar.setup.sql_in
```diff
@@ -46,12 +46,12 @@ CREATE TABLE cifar_10_sample_text_batched AS
 
 -- Insert a new row with NULL as the dependent var (one-hot encoded)
 UPDATE cifar_10_sample_text_batched
-	SET dependent_var = convert_array_to_bytea(ARRAY[0,0,1,0,0]::REAL[]) WHERE buffer_id=0;
+	SET dependent_var = convert_array_to_bytea(ARRAY[0,0,1,0,0]::smallint[]) WHERE buffer_id=0;
```
reductionista: This function requires REAL[] as input, doesn't it?  Seems like we need one for SMALLINT defined also... or does it automatically cast it from SMALLINT to REAL as it runs?

kaknikhil: `convert_array_to_bytea` is overloaded and accepts both real[] and smallint[] as arg types.

PR #452, opened 2019-10-16
DL: Remove quote_ident to allow tables on schemas

orhankislal:
JIRA: MADLIB-1388

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.py_in
@@ -38,7 +38,6 @@ from utilities.utilities import is_platform_pg
 from utilities.utilities import get_segments_per_host
 from utilities.utilities import madlib_version
 from utilities.validate_args import get_expr_type
-from utilities.validate_args import quote_ident
 from utilities.control import MinWarning
 import tensorflow as tf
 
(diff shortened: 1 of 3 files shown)


orhankislal:
I have added a test with a special character in table name. 
I am not sure if we actually have a sql injection risk in this case. The very first case these user-defined strings get used in a sql query is in the table exists function. There, we put the string in quotes before checking against pg_namespace. Since it is treated as a string, not a sql command, I think the check will fail before we execute anything else.

reductionista:
> I have added a test with a special character in table name.
> I am not sure if we actually have a sql injection risk in this case. The very first case these user-defined strings get used in a sql query is in the table exists function. There, we put the string in quotes before checking against pg_namespace. Since it is treated as a string, not a sql command, I think the check will fail before we execute anything else.

Adding quotes manually will allow it to work with table names that have spaces, tabs, or capital letters in their name.  But it won't help guard against SQL injection, which usually works by the attacker ending the quotes and then inserting a command, then starting the quotes again later.   I think this is the main reason for the existence of the `quote_ident` function, since just using quotes would work to handle most other cases.  It also allows table names to have quotes in them, which doesn't work with manual quoting--but that's a fairly obscure use case that doesn't seem important to support.

PR #453, opened 2019-10-21
Add keras version to the docs and release notes

orhankislal:


RELEASE_NOTES
```diff
@@ -9,6 +9,14 @@ commit history located at https://github.com/apache/madlib/commits/master.
 
 Current list of bugs and issues can be found at https://issues.apache.org/jira/browse/MADLIB.
 
+—-------------------------------------------------------------------------
+MADlib v1.17:
+
+Release Date:
+
+Other:
+    - DL: Supported keras version is fixed to 2.2.4
```
kaknikhil: 1. Do we only support one keras version? I think we support 2.2.4 and lower but not sure what's the lowest version we support
1. Same goes for tensorflow, we support 1.14 and lower but not sure about the lowest version we support

src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
```diff
@@ -77,6 +77,9 @@ typically resulting faster and smoother convergence [3].
 You can also do inference on models that have not been trained with MADlib,
 but rather imported from an external source.
 
+Note that the following MADlib functions are targetting a specific Keras
+version (2.2.4). Using a newer or older version may or may not work as intended.
```
kaknikhil: 1. /s/targetting/targeting

PR #468, opened 2020-01-07
DL: Update tests calling trap_error()

khannaekta:
Prior to this PR, due to the missing `SELECT`, call to
`trap_error()` would always return 1, irrespective of whether
the execution of the query would error out or not. This commit
updates such tests to pass in an executable SQL query.
This PR also fixes a bug in predict that was missed due to
the missing `SELECT` in the query param passed to trap_error().

### DIFF
--- a/src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
+++ b/src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
@@ -341,13 +341,14 @@ SELECT assert (dependent_vartype   = 'text' AND
 -- data_preprocessor_input_batch was created for boolean dependent var,
 -- trying to create validation data using this table for reference, but
 -- passing a non-boolean dep var should error out.
-select assert(trap_error($TRAP$validation_preprocessor_dl(
+DROP TABLE IF EXISTS validation_out, validation_out_summary;
+select assert(trap_error($TRAP$SELECT validation_preprocessor_dl(
   'validation_input',
   'validation_out',
   'y_new',
   'x_new',
   'data_preprocessor_input_batch');$TRAP$) = 1,
-    'Passing integer dep var while expecting boolean dep var should error out.');
+    'Passing boolean dep var while expecting text dep var should error out.');
 
 
 -- test double precision type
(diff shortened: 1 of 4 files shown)


src/ports/postgres/modules/deep_learning/test/input_data_preprocessor.sql_in
```diff
@@ -341,7 +341,7 @@ SELECT assert (dependent_vartype   = 'text' AND
 -- data_preprocessor_input_batch was created for boolean dependent var,
 -- trying to create validation data using this table for reference, but
 -- passing a non-boolean dep var should error out.
-select assert(trap_error($TRAP$validation_preprocessor_dl(
+select assert(trap_error($TRAP$SELECT validation_preprocessor_dl(
```
kaknikhil: This trap_error is still failing for the wrong reason i.e. 'the output table `validation_out` already exists'. We need to add the drop table code`DROP TABLE IF EXISTS validation_out, validation_out_summary;` before running the `select validation_preprocessor_dl` statement

src/ports/postgres/modules/deep_learning/test/madlib_keras_predict.sql_in
```diff
@@ -85,7 +85,7 @@ SELECT assert(estimated_y IN (0,1),
 FROM cifar10_predict;
 
 DROP TABLE IF EXISTS cifar10_predict;
-SELECT assert(trap_error($TRAP$madlib_keras_predict(
+SELECT assert(trap_error($TRAP$SELECT madlib_keras_predict(
```
kaknikhil: This is also failing for the wrong reason i.e. wrong column names. 
`id` should be replaced by `buffer_id` and `x` by `independent_var`. 
Also once you make these changes and run this test without the trap_error function, the failure is not meaningful 
```
ERROR:  plpy.SPIError: function array_upper(bytea, integer) does not exist
LINE 2:                 SELECT array_upper(independent_var, 1) AS n_...
                               ^
QUERY:
                SELECT array_upper(independent_var, 1) AS n_0, array_upper(independent_var, 2) AS n_1, array_upper(independent_var, 3) AS n_2
                FROM cifar_10_sample_batched
                LIMIT 1
```
But that should be taken care of in a future PR.

PR #470, opened 2020-01-14
DL: Fix bugs 

khannaekta:
This PR fixes the following bugs:
1. metrics_elapsed_time for fit multi model was not captured correctly in the info table
2. Drop leftover tables

- [X] Add the module name, JIRA# to PR/commit and description.
- [X] Add tests for the change.


src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in
```diff
@@ -230,7 +231,7 @@ class FitMultipleModel():
                 images_per_seg,
```
reductionista: On line 223 (call to compute_loss_and_metrics(), metric_eval_time is an unused local var now, so we could just replace it with _ 

(Unimportant, but  I think that's what we've done most other places.)

fmcquillan99:
LGTM

PR #473, opened 2020-01-22
Release117

orhankislal:

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -10,7 +10,7 @@ commit history located at https://github.com/apache/madlib/commits/master.
 Current list of bugs and issues can be found at https://issues.apache.org/jira/browse/MADLIB.
 
 —-------------------------------------------------------------------------
-MADlib v1.17:
+MADlib v1.17.0:
 
 Release Date: 2020-Mar-31
 
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -12,10 +12,67 @@ Current list of bugs and issues can be found at https://issues.apache.org/jira/b
 —-------------------------------------------------------------------------
 MADlib v1.17:
 
-Release Date:
+Release Date: 2020-Mar-31
+
+New features
+    - DL: Add optional params to madlib_keras_fit_multiple_model (MADLIB-1397)
+    - DL: Fit and evaluate changes for asymmetric cluster config (MADLIB-1393)
+    - DL: Make param search fit() function work with existing evaluate and predict (MADLIB-1387)
+    - DL: ParamSearch: Add utility function for generating model selection table (MADLIB-1375)
+    - DL: Predict changes for asymmetric cluster config (MADLIB-1394)
+    - DL: Preprocessor should evenly distribute data on an arbitrary number of segments (MADLIB-1378)
+    - DL: Preprocessor support for asymmetric segment distribution (MADLIB-1392)
+    - DL: Remove model_arch_table column from the output of load_model_selection_table (MADLIB-1381)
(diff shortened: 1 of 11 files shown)


orhankislal:
jenkins retest this

RELEASE_NOTES
```diff
@@ -12,10 +12,66 @@ Current list of bugs and issues can be found at https://issues.apache.org/jira/b
...
+    - DL: Support schema qualified output tables for fit and fit_multiple (MADLIB-1417)
+    - Graph: APSP fails if both vertex id column and edge src column has the same name (MADLIB-1407)
+    - Graph: ASPS Path Function fails if src or dest column type is bigint (MADLIB-1408)
+    - Graph: Graph/wcc fails if the user specifies a schema for the output table (MADLIB-1411)
+    - Kmeans: k-means related functions must use same default distance function (MADLIB-1383)
+    - LDA: Term frequency and LDA - turn off notices (MADLIB-1395)
+    - MADlib cannot be built on PowerPC machines with Linux (MADLIB-1410)
+    - Pivot:  Pivot documentation should say "out_table" instead of "output_table" (MADLIB-1376)
 
 Other:
```
fmcquillan99: Please add to this section 
```
- DL: CUDA GPU memory cannot be released until the process holding it is terminated.  This process holds the GPU memory until one of the following two things happen: query finishes and user logs out of the Postgres client/session; or, query finishes and user waits for the timeout set by `gp_vmem_idle_resource_timeout`. The default value for this timeout in Greenplum is 18 sec, but it can be changed.
```

orhankislal: Should this go in `big fixes` or `other`?

fmcquillan99: other

src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.sql_in
```diff
@@ -1016,7 +1016,7 @@ name                      |
 description               |
 start_training_time       | 2019-12-16 18:54:33.826414
 end_training_time         | 2019-12-16 18:56:19.106321
-madlib_version            | 1.17-dev
+madlib_version            | 1.17
```
reductionista, line 1019: Maybe these should also say 1.17.0?

Also:  it doesn't show up here in the PR, but in `deep_learning/madlib_keras_fit.sql_in` the versions in the examples still say 1.16

orhankislal:
> I tried the upgrade on my MacBookPro, from 1.16 to 1.17.0 and it threw an exception:
> 
> ```
> ERROR:  column p.prokind does not exist
> LINE 12:                         p.prokind = 'a' AS proisagg
>                                  ^
> HINT:  Perhaps you meant to reference the column "p.probin".
> : ERROR : False
> ```
> 
> This is with postgres 10.5:
> PostgreSQL 10.5 on x86_64-apple-darwin17.7.0, compiled by Apple LLVM version 10.0.0 (clang-1000.11.45.2), 64-bit

The prokind column does not exist in PG10 (it has `proisagg` and `proiswindow` for the same purposes). Since we only support 2 versions of postgres, we dropped PG10 and added PG12 with this release. Compiling etc. should still work for PG10 but the upgrade process does use this particular column.

reductionista:
Upgrade succeeded on my MacBookPro with gpdb5
```
PostgreSQL 8.3.23 (Greenplum Database 5.23.0+dev.10.g50015fed64 build dev) on x86_64-apple-darwin18.7.0, compiled by GCC Apple LLVM version 10.0.1 (clang-1001.0.46.4), 64-bit compiled on Nov  5 2019 08:49:26 (with assert checking)
```
LGTM!

PR #477, opened 2020-02-10
Graph: Fix ambiguous column name in various functions

orhankislal:
JIRA: MADLIB-1407

Some of the graph functions had a bug that surfaced when the vertex id column
had the same name as the edge src or dest columns. This commit fixes the issue
on apsp, bfs, hits and pagerank.

In addition, it adds tests for every graph function.

Closes #

### DIFF
--- a/src/ports/postgres/modules/graph/test/apsp.sql_in
+++ b/src/ports/postgres/modules/graph/test/apsp.sql_in
@@ -97,3 +97,14 @@ SELECT assert(weight = 2, 'Wrong output in graph (APSP)')
 SELECT assert(parent = 2, 'Wrong output in graph (APSP)')
 	FROM out_gr WHERE src = 0 AND "DEST" = 5 AND grp = 1;
 
+DROP TABLE IF EXISTS out, out_summary;
+ALTER TABLE vertex RENAME COLUMN id TO src;
+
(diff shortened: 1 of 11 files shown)


src/ports/postgres/modules/graph/test/apsp.sql_in
```diff
@@ -97,3 +97,14 @@ SELECT assert(weight = 2, 'Wrong output in graph (APSP)')
 SELECT assert(parent = 2, 'Wrong output in graph (APSP)')
 	FROM out_gr WHERE src = 0 AND "DEST" = 5 AND grp = 1;
 
+DROP TABLE IF EXISTS out, out_summary;
```
kaknikhil: Can you add a brief description of the purpose of this test so that it's easier to navigate through the test file ? same goes for all other tests

src/ports/postgres/modules/graph/test/wcc.sql_in
```diff
@@ -158,3 +158,17 @@ SELECT assert(relative_error(num_components, 3) < 0.00001,
         'Weakly Connected Components: Incorrect largest component value.'
     ) FROM count_table WHERE user_id=1;
 
+
+DROP TABLE IF EXISTS out, out_summary;
```
kaknikhil: We should also add the same tests for `graph_vertex_degrees.`

PR #479, opened 2020-02-12
Build: Determine architecture on Linux machines properly

leskin-in:
**Jira**: [MADLIB-1410](https://issues.apache.org/jira/browse/MADLIB-1410)

[Current implementation of `file` output parsing](https://github.com/apache/madlib/blob/fc81374d4d280ab4454150cb126137775e0f7ae6/cmake/Utils.cmake#L71) does not parse the architecture names properly.

`file` command output is kernel-dependent. On Linux, the widest used implementation is http://www.darwinsys.com/file/. The definitions of its outputs for ELF files for the most recent version can be found at https://github.com/file/file/blob/master/magic/Magdir/elf.

This PR implements proper parsing of these outputs and their conversion to Mac OS `lipo` format.

### Notes
Not sure the `lipo` architecture format is the correct choice, but I suppose its values may currently be expected in various places.

I would propose the following:
* Introduce some "unknown" architecture in order not to limit architecture support. Not sure this is possible without extra changes, though.
* Remove [`word_length`](https://github.com/apache/madlib/blob/fc81374d4d280ab4454150cb126137775e0f7ae6/cmake/Utils.cmake#L99) CMake macro. As far as I understand, it is currently not used anywhere.
* Rewrite the architecture recognition CMake macro `architecture` and use `file` both on Mac OS and on Linux. This would simplify the parsing logic of the macro, and it would allow to determine word length directly (`file` output contains it) instead of hardcoding it in `word_length`.

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -68,58 +68,23 @@ macro(get_dir_name OUT_DIR IN_PATH)
     endif(${IN_PATH} MATCHES "^.+/[^/]*\$")
 endmacro(get_dir_name)
 
-macro(architecture FILENAME OUT_ARCHITECTURE)
-    if (APPLE)
-        # On Mac OS X, "lipo" is a more reliable way of finding out the
-        # architecture of an executable
-        osx_archs("${FILENAME}" _ARCHITECTURE)
-    else(APPLE)
-        # On linux, 'file' is used to determine the architecture. However, its
-        # output is kernel-dependent. http://www.darwinsys.com/file/ is the
-        # widest used implementation. Its outputs for most recent version are
-        # defined in
-        # https://github.com/file/file/blob/master/magic/Magdir/elf
-        execute_process(
-            COMMAND file "${FILENAME}"
-            OUTPUT_VARIABLE _FILE_OUTPUT
-            OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-        # Filter out known architectures
-        string(
-            REGEX MATCH
-            "x86-64|Intel 80386|PowerPC or cisco 4500|64-bit PowerPC or cisco 7500"
-            _FILE_OUTPUT_MATCHED "${_FILE_OUTPUT}"
-        )
-        # Convert the result to Mac OS X lipo format
-        if(_FILE_OUTPUT_MATCHED STREQUAL "x86-64")
-            set(_ARCHITECTURE "x86_64")
-        elseif(_FILE_OUTPUT_MATCHED STREQUAL "Intel 80386")
-            set(_ARCHITECTURE "i386")
-        elseif(_FILE_OUTPUT_MATCHED STREQUAL "PowerPC or cisco 4500")
-            set(_ARCHITECTURE "ppc")
-        elseif(_FILE_OUTPUT_MATCHED STREQUAL "64-bit PowerPC or cisco 7500")
-            set(_ARCHITECTURE "ppc64")
-        else()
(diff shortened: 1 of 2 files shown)


cmake/Utils.cmake
```diff
@@ -74,16 +74,34 @@ macro(architecture FILENAME OUT_ARCHITECTURE)
...
         execute_process(
             COMMAND file "${FILENAME}"
             OUTPUT_VARIABLE _FILE_OUTPUT
             OUTPUT_STRIP_TRAILING_WHITESPACE)
 
         # Filter out known architectures
-        string(REGEX MATCHALL "x86[-_]64|i386|ppc|ppc64" _ARCHITECTURE "${_FILE_OUTPUT}")
-
-        # Normalize (e.g., some vendors use x86-64 instead of x86_64)
-        string(REGEX REPLACE "x86[-_]64" "x86_64" _ARCHITECTURE "${_ARCHITECTURE}")
```
orhankislal: Is there a particular reason why we removed this regex replace and comment? I think we should have a similar failsafe and add another `strequal` at line 94.

cmake/Utils.cmake
```diff
@@ -74,16 +74,34 @@ macro(architecture FILENAME OUT_ARCHITECTURE)
...
+        string(
+            REGEX MATCH
+            "x86-64|Intel 80386|PowerPC or cisco 4500|64-bit PowerPC or cisco 7500"
+            _FILE_OUTPUT_MATCHED "${_FILE_OUTPUT}"
+        )
+        # Convert the result to Mac OS X lipo format
+        if(_FILE_OUTPUT_MATCHED STREQUAL "x86-64")
+            set(_ARCHITECTURE "x86_64")
```
orhankislal: Since we are setting the `_ARCHITECTURE` to a single value, do we still need the following if check for the length?

cmake/Utils.cmake
leskin-in: The reason for this is that `file` output cannot be `x86_64` if the http://www.darwinsys.com/file/ implementation is used (according to a statement on their website, it is used by "all known" Linux distributions). In that implementation, `x86-64` was introduced in https://github.com/file/file/commit/e7deb553507b03e6f54829882ae009e648870eea#diff-d55a8317b25513ddc5ca1e726dc3f794R88, and since then always included a dash, not an underscore.

I suppose the regex (which was introduced in https://github.com/apache/madlib/commit/eb933f7cec9ea62a082416910647c7d1e9afcc09#diff-5339764dc1c331c74d366bd06bc5cb58R37) might be an analogy with `lipo` output parsing.

cmake/Utils.cmake
leskin-in: It may be required by the `if(APPLE)` case. The check in question was introduced outside of the if-else construct by https://github.com/apache/madlib/commit/91f8f67aa192b13b87171130523cc4c96dc0b9d2#diff-5339764dc1c331c74d366bd06bc5cb58R22 (for some reason, the check was not placed exclusively in `else(APPLE)` branch).

leskin-in:
The last commit https://github.com/apache/madlib/pull/479/commits/5fd815f2eb47ed15e0764f01c3178b73156d40fe is another way the architecture could be determined. `uname` produces shorter (machine-readable) and probably more consistent output.

This change leads to substantial simplification of the `architecture()` CMake macro.

However, this method to determine architecture is fundamentally different from the previously used one: `file` checks the actual Postgres / Greenplum binary, while `uname` outputs the architecture of the machine the build is run on. If the build is run on a machine whose architecture is different from the target one, `architecture()` would return incorrect results.

PR #480, opened 2020-02-19
Summary: Fix defaults when the user passes NULL

orhankislal:
JIRA: MADLIB-1413

Summarizer used python defaults for setting up the optional parameters.
If the user passes a NULL, these values get overwritten. This commit
fixes this issue by moving the defaults inside the code.

### DIFF
--- a/src/ports/postgres/modules/summary/Summarizer.py_in
+++ b/src/ports/postgres/modules/summary/Summarizer.py_in
@@ -11,8 +11,8 @@ class Summarizer:
 
     def __init__(self, schema_madlib, source_table, output_table,
                  target_cols, grouping_cols, distinctify, get_quartiles,
-                 xtileify='Exact', ntile_array=None, how_many_mfv=10,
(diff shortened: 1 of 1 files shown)


fmcquillan99:
```
SELECT madlib.summary (                                                                                                                         'abalone_encoded',  -- source_table                                                                                                                  'abalone_summary_exact',  -- output_table                                                                                                            NULL,  -- target_cols                                                                                                                                NULL,  -- grouping_cols                                                                                                                              TRUE,  -- get_distinct                                                                                                                               TRUE,  -- get_quartiles                                                                                                                              NULL,  -- quantile_array                                                                                                                             20,    -- how_many_mfv                                                                                                                               TRUE -- get_estimate                                                                                                                                     );
                   summary                    
----------------------------------------------
 (abalone_summary_exact,14,0.682248830795288)
(1 row)

Time: 829.001 ms
```
LGTM

PR #482, opened 2020-02-21
Added support for quoted identifiers for install-check

ferrarius27:
For database names containing  arbitrary sequence of characters e.g. TEST"ME,TEST ME,etc.

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -674,7 +674,6 @@ def _process_py_sql_files_in_modules(modset, args_dict):
 
         # Loop through all SQL files for this module
         source_files = glob.glob(mask)
-        source_files = [s for s in source_files if '.setup' not in s]
         if calling_operation == INSTALL_DEV_CHECK and madpack_cmd != 'install-check':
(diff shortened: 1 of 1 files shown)


src/madpack/madpack.py
```diff
@@ -674,7 +674,6 @@ def _process_py_sql_files_in_modules(modset, args_dict):
 
         # Loop through all SQL files for this module
         source_files = glob.glob(mask)
-        source_files = [s for s in source_files if '.setup' not in s]
```
orhankislal: This line was included to ensure the common setup files for tests (like `madlib_keras_cifar.setup.sql_in`) are not treated as individual tests.

ferrarius27: Sorry for the late response,looks like this line was accidentally deleted,it has nothing to do with the fix

orhankislal: No worries, I removed this line before merging the code. Thanks for the PR.

ferrarius27: Ok,thanks.

PR #483, opened 2020-02-25
Graph: Fix column type and output table schema bugs

orhankislal:
This PR includes 2 commits to fix MADLIB-1408 and MADLIB-1411

### DIFF
--- a/src/ports/postgres/modules/graph/apsp.sql_in
+++ b/src/ports/postgres/modules/graph/apsp.sql_in
@@ -78,7 +78,7 @@ contain the column specified in the 'vertex_id' parameter below.</dd>
 
 <dt>vertex_id</dt>
 <dd>TEXT, default = 'id'. Name of the column in 'vertex_table' containing
-vertex ids.  The vertex ids are of type INTEGER with no duplicates.
+vertex ids.  The vertex ids are of type BIGINT with no duplicates.
 They do not need to be contiguous.</dd>
 
 <dt>edge_table</dt>
@@ -90,9 +90,9 @@ Column naming convention is described below in the 'edge_args' parameter.</dd>
 <dd>TEXT. A comma-delimited string containing multiple named arguments of
 the form "name=value". The following parameters are supported for
 this string argument:
-  - src (INTEGER): Name of the column containing the source vertex ids in the
+  - src (BIGINT): Name of the column containing the source vertex ids in the
   edge table. Default column name is 'src'.
-  - dest (INTEGER): Name of the column containing the destination vertex ids
+  - dest (BIGINT): Name of the column containing the destination vertex ids
   in the edge table. Default column name is 'dest'.
   - weight (FLOAT8): Name of the column containing the edge weights in the
   edge table. Default column name is 'weight'.</dd>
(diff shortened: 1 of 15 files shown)


src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -411,6 +412,24 @@ def graph_sssp_get_path(schema_madlib, sssp_table, dest_vertex, path_table,
...
+        default_args = {'src': 'src', 'dest': 'dest', 'weight': 'weight'}
+        edge_params = extract_keyvalue_params(edge_args,
+                                              params_types,
+                                              default_args)
+        src = edge_params["src"]
+        dest = edge_params["dest"]
+        weight = edge_params["weight"]
+
+        # Find the appropriate column type for correct concat operations
+        types = get_cols_and_types(sssp_table)
```
kaknikhil:
```suggestion
        names_and_types = get_cols_and_types(sssp_table)
```

src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -411,6 +412,24 @@ def graph_sssp_get_path(schema_madlib, sssp_table, dest_vertex, path_table,
...
+        final_type = 'integer'
+        for i in types:
```
kaknikhil:
```suggestion
        for name,type in names_and_types:
            # We want to check the column types except the weight column
            if type == 'bigint'  and name != weight:
                final_type = 'bigint'
```

src/ports/postgres/modules/graph/apsp.sql_in
```diff
@@ -412,9 +412,9 @@ m4_ifdef(`\_\_HAS_FUNCTION_PROPERTIES\_\_', `MODIFIES SQL DATA', `');
 
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.graph_apsp_get_path(
     apsp_table             TEXT,
-    source_vertex          INT,
-  dest_vertex            INT,
-  path_table             TEXT
+    source_vertex          BIGINT,
```
kaknikhil: will this and other such changes be backwards compatible ?

src/ports/postgres/modules/graph/wcc.py_in
```diff
@@ -294,7 +294,7 @@ def wcc(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
     if not is_platform_pg():
         # Drop intermediate table created for Greenplum
         plpy.execute("DROP TABLE IF EXISTS {0}".format(edge_inverse))
-    plpy.execute("ALTER TABLE {0} RENAME TO {1}".format(newupdate, out_table))
+    plpy.execute("CREATE TABLE {0} AS SELECT * FROM {1}".format(out_table, newupdate))
```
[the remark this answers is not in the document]
orhankislal: `newupdate` is dropped at the end of the function(line 316-317). We need this change because postgresql doesn't allow altering the schema of a table. If the user wants the output table in a specific schema, alter table line will fail.

src/ports/postgres/modules/graph/apsp.sql_in
orhankislal: Yes, we still have tests that pass integer columns to these functions that work as expected.

PR #484, opened 2020-02-26
MFVSketch: Remove duplicate results on GPDB

orhankislal:
JIRA: MADLIB-1412

mfvsketch merge function did not filter out duplicate entries if
a value is in the mfv list for multiple segments.

This commit fixes the issue by keeping a list of added values and
filtering out duplicate entries.

We also noticed the use of memcmp() for comparing two Datums to
check for equality.  There is a note in the postgres/gpdb source
saying this is unsafe, so we converted it over to calling the
postgres function datumIsEqual() which can handle either
little-endian or big-endian byte order, making it more portable.

Closes #


methods/sketch/src/pg_gp/sql/mfv.sql_in
```diff
@@ -22,3 +22,4207 @@ select mfvsketch_top_histogram(NULL::bytea,5) from generate_series(1,100);
...
+(1153,'M',0.59,0.47,0.15,0.9955,0.481,0.232,0.24,8),
+(1154,'F',0.6,0.475,0.16,1.0265,0.485,0.2495,0.2565,9),
+(1155,'M',0.6,0.455,0.17,1.1915,0.696,0.2395,0.24,8),
+(1156,'F',0.6,0.465,0.15,1.1025,0.5455,0.262,0.25,8),
+(1157,'M',0.6,0.465,0.155,1.0165,0.512,0.2465,0.225,10),
+(1158,'F',0.605,0.47,0.165,1.1775,0.611,0.2275,0.292,9),
+(1159,'M',0.605,0.475,0.14,1.1175,0.555,0.257,0.274,9),
+(1160,'M',0.605,0.48,0.17,1.1835,0.582,0.2365,0.317,10),
+(1161,'F',0.605,0.475,0.165,1.056,0.433,0.2195,0.357,9),
+(1162,'M',0.61,0.485,0.16,1.014
```
kaknikhil: Can we also add a test for `mfvsketch_top_histogram` ?

methods/sketch/src/pg_gp/mfvsketch.c
```diff
@@ -568,35 +572,77 @@ bytea *mfvsketch_merge_c(bytea *transblob1, bytea *transblob2)
...
+                    cnt++;
+                } else {
+                    elog(WARNING, "Merge received more than %d values from transition functions", 2*transval2->max_mfvs);
+                }
+            }
+            j++;
         }
     }
+
+    if (!transval1->typByVal){
```
kaknikhil: Why do we have this check here ? Is it possible that `transval1->typByVal` is null but values_seen has also been allocated ?

kaknikhil:
I would also suggest separating out the two fixes into their own commits with a detailed commit message.

methods/sketch/src/pg_gp/sql/mfv.sql_in
orhankislal: This issue doesn't apply to `mfvsketch_top_histogram` since it does not have a merge function.

methods/sketch/src/pg_gp/mfvsketch.c
orhankislal: `transval1->typByVal` is a boolean and set to `true` or `false` during initialization (`mfv_init_transval`) by a db function (`get_typbyval`). We have this check here because we want to free the Datums if they were passed by reference. Pass by value Datums are like primitive types and they are taken care of once the array is freed at line 645

PR #485, opened 2020-03-03
DL: Remove fit final multiple and improve serialization of image_count+weights

kaknikhil:
1. Based on our experiments, we noticed that at the end of each hop, fit
multiple would start using a lot of memory (from 80 GB to ~ 140 GB on
the master host of our gcp cluster) and we attributed that memory spike
to the final function.
Removing the final function brought down the memory to around 30-40 GB
and also made the places10 query 3 times faster

2. Improvements to "serializing weights with image count"

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.sql_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.sql_in
@@ -1492,15 +1492,6 @@ PythonFunctionBodyOnlyNoSchema(`deep_learning', `madlib_keras')
 $$ LANGUAGE plpythonu
 m4_ifdef(`__HAS_FUNCTION_PROPERTIES__', `NO SQL', `');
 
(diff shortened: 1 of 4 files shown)


src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.sql_in
```diff
@@ -1537,6 +1537,5 @@ CREATE AGGREGATE MADLIB_SCHEMA.fit_step_multiple_model(
     /* is_final_iteration */         BOOLEAN
 )(
     STYPE=BYTEA,
-    SFUNC=MADLIB_SCHEMA.fit_transition_multiple_model,
-    FINALFUNC=MADLIB_SCHEMA.fit_final_multiple_model
```
orhankislal: It seems the sql and python functions for `fit_final_multiple_model` are still in code even though we don't call them anymore. Is there a particular reason for keeping them?

kaknikhil: No reason, I will remove them.

PR #486, opened 2020-03-06
Graph: Filter out infinite paths

orhankislal:
JIRA: MADLIB-1415

SSSP, APSP and BFS output tables had entries for unconnected vertices,
denoted by a NULL parent column. This commit filters these rows to
clean up the output tables.

### DIFF
--- a/src/ports/postgres/modules/graph/measures.py_in
+++ b/src/ports/postgres/modules/graph/measures.py_in
@@ -181,13 +181,7 @@ class Graph(object):
             CREATE TABLE {out_table} AS
             SELECT
                 {grouping_cols_comma}
-                -- Filtering 'Infinity' occurs in CASE instead of WHERE clause
-                -- so that the edge is part of the average value i.e. part of
-                -- the count of paths but zero addition to sum of distances.
-                AVG(CASE WHEN {e.weight} = 'Infinity'::double precision
-                            THEN 0::double precision
-                            ELSE {e.weight}::double precision
-                    END) as avg_path_length
+                AVG({e.weight}::double precision) as avg_path_length
             FROM {apsp_table}
             WHERE {e.src} != {e.dest}
                   {filter_clause}
--- a/src/ports/postgres/modules/graph/test/measures.sql_in
+++ b/src/ports/postgres/modules/graph/test/measures.sql_in
@@ -86,7 +86,7 @@ SELECT assert(diameter=14, 'Invalid value for diameter') FROM public.__madlib__o
 DROP TABLE IF EXISTS public.__madlib__out_avg_path_length;
 SELECT graph_avg_path_length('out_apsp', 'public.__madlib__out_avg_path_length');
 SELECT * FROM public.__madlib__out_avg_path_length;
-SELECT assert(relative_error(avg_path_length, 2.0178) < 1e-2,
+SELECT assert(relative_error(avg_path_length, 2.97) < 1e-2,
               'Invalid value for avg_path_length') FROM public.__madlib__out_avg_path_length;
 
 -- Compute the in and out degrees
--- a/src/ports/postgres/modules/graph/test/measures.sql_in
+++ b/src/ports/postgres/modules/graph/test/measures.sql_in
@@ -64,51 +64,46 @@ SELECT graph_apsp('vertex',      -- Vertex table
                   'out_apsp');   -- Output table of shortest paths
 
 -- Compute the closeness measure for all nodes:
-DROP TABLE IF EXISTS public.__madlib__out_closeness;
-SELECT graph_closeness('out_apsp', 'public.__madlib__out_closeness');
-SELECT * FROM public.__madlib__out_closeness;
+SELECT graph_closeness('out_apsp', 'pg_temp.__madlib__out_closeness');
+SELECT * FROM pg_temp.__madlib__out_closeness;
 
 SELECT assert(relative_error(inverse_sum_dist, 0.04347) < 1e-2 and
               relative_error(inverse_avg_dist, 0.3043) < 1e-2 and
               relative_error(sum_inverse_dist, 3.6833) < 1e-2 and
               k_degree = 7,
               'Incorrect value for closeness')
-FROM public.__madlib__out_closeness
+FROM pg_temp.__madlib__out_closeness
 WHERE src_id = 0;
 
 -- Compute the diameter measure for graph
-DROP TABLE IF EXISTS public.__madlib__out_diameter;
-SELECT graph_diameter('out_apsp', 'public.__madlib__out_diameter');
-SELECT * FROM public.__madlib__out_diameter;
-SELECT assert(diameter=14, 'Invalid value for diameter') FROM public.__madlib__out_diameter;
+SELECT graph_diameter('out_apsp', 'pg_temp.__madlib__out_diameter');
+SELECT * FROM pg_temp.__madlib__out_diameter;
+SELECT assert(diameter=14, 'Invalid value for diameter') FROM pg_temp.__madlib__out_diameter;
 
 -- Compute the average path length measure for graph
-DROP TABLE IF EXISTS public.__madlib__out_avg_path_length;
-SELECT graph_avg_path_length('out_apsp', 'public.__madlib__out_avg_path_length');
-SELECT * FROM public.__madlib__out_avg_path_length;
+SELECT graph_avg_path_length('out_apsp', 'pg_temp.__madlib__out_avg_path_length');
+SELECT * FROM pg_temp.__madlib__out_avg_path_length;
 SELECT assert(relative_error(avg_path_length, 2.97) < 1e-2,
-              'Invalid value for avg_path_length') FROM public.__madlib__out_avg_path_length;
+              'Invalid value for avg_path_length') FROM pg_temp.__madlib__out_avg_path_length;
 
 -- Compute the in and out degrees
-DROP TABLE IF EXISTS public.__madlib__out_degrees;
 SELECT graph_vertex_degrees('vertex',      -- Vertex table
                      'id',          -- Vertix id column (NULL means use default naming)
                      '"EDGE"',        -- "EDGE" table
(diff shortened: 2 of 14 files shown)


fmcquillan99:
(1)
SSSP - filter out infinite paths

```
DROP TABLE IF EXISTS vertex, edge;

CREATE TABLE vertex(
        id INTEGER
        );

CREATE TABLE edge(
        src INTEGER,
        dest INTEGER,
        weight FLOAT8
        );

INSERT INTO vertex VALUES
(0),
(1),
(2),
(3),
(4),
(5),
(6),
(7);

INSERT INTO edge VALUES
(0, 1, 1.0),
(0, 2, 1.0),
(0, 4, 10.0),
(1, 2, 2.0),
(1, 3, 10.0),
(2, 3, 1.0),
(2, 5, 1.0),
(2, 6, 3.0),
(3, 0, 1.0),
(4, 0, -2.0),
(5, 6, 1.0);
```
Note that vertex 7 is disconnect from the rest of the graph ^^^

```
DROP TABLE IF EXISTS out, out_summary;

SELECT madlib.graph_sssp(
                         'vertex',      -- Vertex table
                         NULL,          -- Vertex id column (NULL means use default naming)
                         'edge',        -- Edge table
                         NULL,          -- Edge arguments (NULL means use default naming)
                         0,             -- Source vertex for path calculation
                         'out');        -- Output table of shortest paths

SELECT * FROM out ORDER BY id;
 id | weight | parent 
----+--------+--------
  0 |      0 |      0
  1 |      1 |      0
  2 |      1 |      0
  3 |      2 |      2
  4 |     10 |      0
  5 |      2 |      2
  6 |      3 |      5
(7 rows)
```
Note that `id=7` has no row ^^^
OK

Now check path to 7:
```
DROP TABLE IF EXISTS out_path;
SELECT madlib.graph_sssp_get_path('out',7, 'out_path');

ERROR:  plpy.Error: Graph SSSP: Vertex 7 is not present in the SSSP table out (plpython.c:5038)
  PL/Python function "graph_sssp_get_path", line 21, in <module>
    return sssp.graph_sssp_get_path(**globals())
  PL/Python function "graph_sssp_get_path", line 507, in graph_sssp_get_path
PL/Python function "graph_sssp_get_path"
```
OK


(2)
APSP - filter out infinite paths

```
DROP TABLE IF EXISTS out, out_summary;

SELECT madlib.graph_apsp(
                         'vertex',      -- Vertex table
                         NULL,          -- Vertix id column (NULL means use default naming)
                         'edge',        -- Edge table
                         NULL,          -- Edge arguments (NULL means use default naming)
                         'out');        -- Output table of shortest paths

select * from out order by src, dest;
 src | dest | weight | parent 
-----+------+--------+--------
   0 |    0 |      0 |      0
   0 |    1 |      1 |      1
   0 |    2 |      1 |      2
   0 |    3 |      2 |      2
   0 |    4 |     10 |      4
   0 |    5 |      2 |      2
   0 |    6 |      3 |      5
   1 |    0 |      4 |      3
   1 |    1 |      0 |      1
   1 |    2 |      2 |      2
   1 |    3 |      3 |      2
   1 |    4 |     14 |      0
   1 |    5 |      3 |      2
   1 |    6 |      4 |      5
   2 |    0 |      2 |      3
   2 |    1 |      3 |      0
   2 |    2 |      0 |      2
   2 |    3 |      1 |      3
   2 |    4 |     12 |      0
   2 |    5 |      1 |      5
   2 |    6 |      2 |      5
   3 |    0 |      1 |      0
   3 |    1 |      2 |      0
   3 |    2 |      2 |      0
   3 |    3 |      0 |      3
   3 |    4 |     11 |      0
   3 |    5 |      3 |      2
   3 |    6 |      4 |      5
   4 |    0 |     -2 |      0
   4 |    1 |     -1 |      0
   4 |    2 |     -1 |      0
   4 |    3 |      0 |      2
   4 |    4 |      0 |      4
   4 |    5 |      0 |      2
   4 |    6 |      1 |      5
   5 |    5 |      0 |      5
   5 |    6 |      1 |      6
   6 |    6 |      0 |      6
(38 rows)
```
Note that `id=7` has no row ^^^
OK

Now try to get path to 7:
```
DROP TABLE IF EXISTS out_path;
SELECT madlib.graph_apsp_get_path('out',0,7,'out_path');
ERROR:  plpy.Error: Graph APSP: Vertex 0 and/or 7 is not present in the APSP table 7 (plpython.c:5038)
  PL/Python function "graph_apsp_get_path", line 21, in <module>
    return apsp.graph_apsp_get_path(**globals())
  PL/Python function "graph_apsp_get_path", line 533, in graph_apsp_get_path
PL/Python function "graph_apsp_get_path"
```
OK


(3)
Make BFS output match SSSP and APSP

```
DROP TABLE IF EXISTS vertex, edge;

CREATE TABLE vertex(
        id INTEGER
        );

CREATE TABLE edge(
        src INTEGER,
        dest INTEGER
        );

INSERT INTO vertex VALUES
(0),
(1),
(2),
(3),
(4),
(5),
(6),
(7),
(8),
(9),
(10),
(11);

INSERT INTO edge VALUES
(0, 5),
(1, 0),
(1, 3),
(2, 6),
(3, 4),
(3, 5),
(4, 2),
(8, 9),
(9, 10),
(9, 11),
(10, 8);

DROP TABLE IF EXISTS out, out_summary;

SELECT madlib.graph_bfs(
                         'vertex',      -- Vertex table
                         NULL,          -- Vertix id column (NULL means use default naming)
                         'edge',        -- Edge table
                         NULL,          -- Edge arguments (NULL means use default naming)
                         3,             -- Source vertex for BFS
                         'out');        -- Output table of nodes reachable from source_vertex
                                        -- Default values used for the other arguments

SELECT * FROM out ORDER BY dist,id;
 id | dist | parent 
----+------+--------
  3 |    0 |      3
  1 |    1 |      3
  4 |    1 |      3
  5 |    1 |      3
  0 |    2 |      1
  2 |    2 |      4
  6 |    3 |      2
(7 rows)
```
Note BFS now has output similar to SSSP & APSP with the "self" row included with `dist=0` ^^^

OK

src/ports/postgres/modules/graph/measures.py_in
```diff
@@ -184,10 +184,7 @@ class Graph(object):
                 -- Filtering 'Infinity' occurs in CASE instead of WHERE clause
                 -- so that the edge is part of the average value i.e. part of
                 -- the count of paths but zero addition to sum of distances.
-                AVG(CASE WHEN {e.weight} = 'Infinity'::double precision
-                            THEN 0::double precision
-                            ELSE {e.weight}::double precision
-                    END) as avg_path_length
+                AVG({e.weight}::double precision) as avg_path_length
```
reductionista: We should remove the comment above, which still references the CASE block that's no longer there.

Also wondering:

As I understand it, the original author thought it was important for infinite paths to be included in the denominator of the average value calculation (avg = sum(weights) / count(weights) ) but not in the numerator.  That doesn't seem right to me, but the fact they went to the trouble of specifically pointing it out in the comment makes me wonder if there was some reason for it that we're missing. 

Since we are changing the way in which average path length is calculated, affecting the result that gets returned (not just omitting infinite rows from the table), I just want to make sure we're doing this intentionally.  The average path length returned previously would have been a smaller number than it will be now for disconnected graphs, right?

src/ports/postgres/modules/graph/test/measures.sql_in
```diff
@@ -86,7 +86,7 @@ SELECT assert(diameter=14, 'Invalid value for diameter') FROM public.__madlib__o
 DROP TABLE IF EXISTS public.__madlib__out_avg_path_length;
 SELECT graph_avg_path_length('out_apsp', 'public.__madlib__out_avg_path_length');
 SELECT * FROM public.__madlib__out_avg_path_length;
-SELECT assert(relative_error(avg_path_length, 2.0178) < 1e-2,
+SELECT assert(relative_error(avg_path_length, 2.97) < 1e-2,
```
reductionista: I guess this may answer my question about avg path length--leaving it there just in case.

src/ports/postgres/modules/graph/measures.py_in
fmcquillan99: @njayaram2 Wondering if you recall this and and can provide some guidance here?  thanks

orhankislal: It seems the average path length is usually defined for connected graphs. If a graph is unconnected, the average path length becomes infinity as well. There are suggestions like using the largest connected component's avg path length (which is possible to do with MADlib wcc function) [https://mathinsight.org/definition/network_mean_path_length](url).

I think the previous way of calculating was plain wrong. With this commit, the average length of non-infinite paths is calculated, which may or may not be useful. If we want to specifically limit this function to connected graphs, we can do that as well.

reductionista: Sounds good to me!

Also backing up this decision, I just noticed this:
```
postgres=# select madlib.graph_avg_path_length();
                             graph_avg_path_length
-------------------------------------------------------------------------------

 This function computes the average of the shortest paths between each pair of
 vertices. Average path length is based on "reachable target vertices", so it
 ignores infinite-length paths between vertices that are not connected.

 For more details:
     SELECT madlib.graph_avg_path_length('usage')

(1 row)
```
Pretty clear the previous behavior was a bug!  Glad we've fixed it.

fmcquillan99: If you have, say, n disconnected sub-graphs of roughly the same size, this calc is still OK, right?  i.e., it just drops vertex pairs that are not connected (infinite dist) and would calculate average path length based on pairwise distances that are not infinite.

orhankislal: Yes. The number and/or size of components are irrelevant, we just look at every non-infinite path and take the average.

fmcquillan99: I see.  In that case I think this PR is ready to merge.  Thanks @reductionista for asking an interesting question to start, and to both or you for sorting it out.

PR #489, opened 2020-03-12
DL: Set plan_cache_mode to choose generic plan when passing weights for GPDB6

khannaekta:
JIRA: MADLIB-1414

This PR reverts the previous commit for avoiding constant folding of weights in GPDB6 plans. Instead, there is a GUC introduced in GPDB6(`plan_cache_mode`) that allows controlling whether to choose generic plan(not constant folding the passed in params)/custom plans(constant folding the passed in params) for execution. This PR uses this GUC for choosing generic plan for GPDB6 execution.
This PR also fixes a bug in the way `class_values` were populated in fit multiple's summary table; it would fail when the class_values had `NULL` values in the array.

NOTE: For the prior versions of GPDB6, where this GUC doesn't exist, the DL queries will fail with `memory alloc` error for those versions.  

- [X] Add the module name, JIRA# to PR/commit and description.
- [X] Add tests for the change.

### DIFF
--- a/src/ports/postgres/modules/convex/mlp_igd.py_in
+++ b/src/ports/postgres/modules/convex/mlp_igd.py_in
@@ -1051,8 +1051,8 @@ def mlp_predict(schema_madlib, model_table, data_table, id_col_name,
         else:
             intermediate_col = unique_string()
             if classes:
-                score_format = create_cols_from_array_sql_string(
-                    classes, intermediate_col, 'prob', 'double precision', False, 'MLP') 
+                score_format, _ = create_cols_from_array_sql_string(
+                    classes, intermediate_col, 'prob', 'double precision', False, 'MLP')
             else:
                 # Case when the training step did not have to one-hot encode
                 # the dependent var.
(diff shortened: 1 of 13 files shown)


orhankislal:
The PR seems fine but I have one question. I tried running dev check on GPDB 6.4 (On OS X) and it didn't fail. Since this version of greenplum doesn't have the guc yet, isn't it supposed to fail at some point?

khannaekta:
> The PR seems fine but I have one question. I tried running dev check on GPDB 6.4 (On OS X) and it didn't fail. Since this version of greenplum doesn't have the guc yet, isn't it supposed to fail at some point?

dev-check, doesn't fail since we do not pass in large models that will have large weights(~600MB). Didn't add those tests to avoid slowing down the dev-check.

orhankislal:
> > The PR seems fine but I have one question. I tried running dev check on GPDB 6.4 (On OS X) and it didn't fail. Since this version of greenplum doesn't have the guc yet, isn't it supposed to fail at some point?
> 
> dev-check, doesn't fail since we do not pass in large models that will have large weights(~600MB). Didn't add those tests to avoid slowing down the dev-check.

OK that makes sense. I just wanted to clarify that in the absence of the guc some models can still be built.

PR #491, opened 2020-03-23
 DL: Fix disk issue by using truncate guc 

kaknikhil:
While testing places10 with fit multiple (gpdb5 and gpdb6, 10 iterations and 20 msts), we ran out of disk space although we had at least 1.5T left at the beginning of the query.

Fixed this by using a gpdb (> 6.4) only guc that allows for releasing disk space inside a sub transaction by calling truncate (see commit message for more details)

Other changes (see individual commit for more details)
1. Utilities: Improve rename table logic 
1. Graph: Add select statements to graph tests 
1. DL: add tests for asserting that guc values are unchanged 
1. Use unlogged tables for performance improvements 

JIRA: MADLIB-1406


src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in
```diff
@@ -484,7 +545,7 @@ class FitMultipleModel():
             if self.validation_table:
                 self.update_info_table(mst, False)
 
-    def run_training(self):
+    def run_training(self, mst_idx):
```
orhankislal: It looks like we don't use this new parameter inside the function.

src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in
```diff
@@ -185,13 +226,16 @@ class FitMultipleModel():
                 mst_row = [self.grand_schedule[dist_key][mst_idx]
                            for dist_key in self.dist_keys]
                 self.create_mst_schedule_table(mst_row)
+                self.is_final_training_call = (iter == self.num_iterations and mst_idx == total_msts-1)
                 if mst_idx == 0:
                     start_iteration = time.time()
-                self.run_training()
+                self.run_training(mst_idx)
```
orhankislal: Same question as to the function definition. Is `mst_idx` being used during `run_training`?

src/ports/postgres/modules/utilities/utilities.sql_in
```diff
@@ -541,3 +541,49 @@ BEGIN
     RETURN 0;
 END;
 $$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.is_ver_greater_than_gp_640_or_pg_11()
```
orhankislal: I understand why we need this very specific function but it would be great to add a few lines of comments for context. It is especially confusing because the function is in utilities and I usually expect them to be somewhat generic. Please note that I am not suggesting to move the function to DL, just some comments.

src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in
kaknikhil: mst_idx has been added because it helps us print per hop time when we need to debug the fit multiple query. We might also add a verbose flag in the future to print per hop time.

src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in
kaknikhil: see https://github.com/apache/madlib/pull/491#discussion_r396854057

src/ports/postgres/modules/utilities/utilities.sql_in
kaknikhil: makes sense. will do

src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.sql_in
```diff
@@ -88,6 +88,10 @@ You can set up the models and hyperparameters to try with the
 Model Selection</a> utility to define the unique combinations
 of model architectures, compile and fit parameters.
 
+@note If madlib_keras_fit_multiple_model is running on gpdb5, the database will
+keep on adding to the disk space(depending on the model size) and will only
+release the disk space once the fit multiple query has completed execution.
+
```
fmcquillan99: Suggested slight changes:

@note If 'madlib_keras_fit_multiple_model()' is running on GPDB 5, the database will
keep adding to the disk space (in proportion to model size) and will only
release the disk space once the fit multiple query has completed execution.
This is not the case for GPDB 6+ where disk space is released during the 
fit multiple query.

kaknikhil: sounds good

PR #494, opened 2020-04-14
Update version to 1.18.0-dev

orhankislal:

### DIFF
--- a/deploy/DEB/postinst
+++ b/deploy/DEB/postinst
@@ -22,7 +22,7 @@
 # Source debconf library.
 . /usr/share/debconf/confmodule
 
-MADLIB_VERSION="1.17.0"
+MADLIB_VERSION="1.18.0-dev"
 MADLIB_INSTALL_PATH="InstallPathNotFound"
 
 # Fetching configuration from debconf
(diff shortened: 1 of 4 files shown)


orhankislal:
jenkins retest this please

kaknikhil:
These changes LGTM.

But as soon as we push this commit, the madlib_keras.sql_in and madlib_keras_fit_multiple_model.sql_in dev check tests will start failing because they hardcode the version no. We should instead test that the madlib_version column is not null instead of a specific version.  
See https://github.com/apache/madlib/commit/186a28b6380bd44a87f5240ff9aabd836e023594 for more details

orhankislal:
> These changes LGTM.
> 
> But as soon as we push this commit, the madlib_keras.sql_in and madlib_keras_fit_multiple_model.sql_in dev check tests will start failing because they hardcode the version no. We should instead test that the madlib_version column is not null instead of a specific version.
> See [186a28b](https://github.com/apache/madlib/commit/186a28b6380bd44a87f5240ff9aabd836e023594) for more details

Those are documentation strings, tests do check for not null.

PR #495, opened 2020-04-23
Simplify upgrade

orhankislal:

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -518,40 +518,34 @@ def _db_upgrade(schema, filehandle, db_madlib_ver):
         info_(this, vd.get_dependency_graph_str(), True)
         info_(this, "*" * 50, True)
 
-        c_udf = ch.get_udf_signature()
         d_udf = vd.get_depended_func_signature('UDF')
-        cd_udf = [udf for udf in d_udf if udf in c_udf]
-        if len(cd_udf) > 0:
+        if len(d_udf) > 0:
             error_(this, """
                 User has objects dependent on following updated MADlib functions!
                     {0}
-                These objects will fail to work with the updated functions and
+                These objects might fail to work with the updated functions and
                 need to be dropped before starting upgrade again.
-                """.format('\n\t\t\t\t\t'.join(cd_udf)), False)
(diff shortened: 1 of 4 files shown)


src/madpack/madpack.py
```diff
@@ -1200,10 +1194,63 @@ def create_install_madlib_sqlfile(args, madpack_cmd):
...
+          CASE WHEN o.oprkind='r' THEN NULL
+            ELSE pg_catalog.format_type(o.oprright, NULL) END AS "right_op"
+        FROM pg_catalog.pg_operator o
+             LEFT JOIN pg_catalog.pg_namespace n ON n.oid = o.oprnamespace
+        WHERE n.nspname OPERATOR(pg_catalog.~) '^(%s)$'
+        ORDER BY 1, 2, 3, 4;
+        """ % schema.lower(), True)
+
+    # Drop the Operator Class before the individual operators
+    drop_str = "DROP OPERATOR CLASS IF EXISTS {0}.svec_ops USING btree; \n".format(schema)
```
khannaekta, line 1247: Shouldn't this go through `self._udoc` instead of just dropping the `svec_ops`? Also, do we still need to call `drop_changed_udoc()` at [L566](https://github.com/apache/madlib/pull/495/files#diff-aee20fa9b7d6f9749f7683e14158369aR566)

src/madpack/madpack.py
```diff
@@ -565,13 +559,13 @@ def _db_upgrade(schema, filehandle, db_madlib_ver):
     sc = uu.ScriptCleaner(schema, portid, con_args, ch)
     info_(this, "Script Cleaner initialized ...", False)
 
-    ch.drop_changed_uda()
+    function_drop_str = get_madlib_function_drop_str(schema)
```
khannaekta, line 562: Is there a specific reason for creating SQL queries and then running them, instead of calling the original functions (`drop_changed_uda()` etc) ?

orhankislal: Those are for dropping the functions, etc. from changelist. Now we use a different query to get all madlib functions.

src/madpack/madpack.py
orhankislal: self._udoc is a part of the changehandler class. It is not available in this context.

PR #496, opened 2020-05-04
DBSCAN: Add new module DBSCAN

orhankislal:
This commit adds the DBSCAN module. It includes the brute force
implementation of DBSCAN as well as a prediction function. In
addition their associated tests and online documentation are
included.

### DIFF
--- a/src/ports/postgres/modules/dbscan/dbscan.py_in
+++ b/src/ports/postgres/modules/dbscan/dbscan.py_in
@@ -1,0 +1,334 @@
+# coding=utf-8
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import plpy
+
+from utilities.control import MinWarning
+from utilities.utilities import _assert
+from utilities.utilities import unique_string
+from utilities.utilities import add_postfix
+from utilities.utilities import NUMERIC, ONLY_ARRAY
+from utilities.utilities import is_valid_psql_type
+from utilities.utilities import is_platform_pg
+from utilities.validate_args import input_tbl_valid, output_tbl_valid
+from utilities.validate_args import is_var_valid
+from utilities.validate_args import cols_in_tbl_valid
+from utilities.validate_args import get_expr_type
+from utilities.validate_args import get_algorithm_name
+from graph.wcc import wcc
+
+BRUTE_FORCE = 'brute_force'
+KD_TREE = 'kd_tree'
+
+def dbscan(schema_madlib, source_table, output_table, id_column, expr_point, eps, min_samples, metric, algorithm, **kwargs):
+
+    with MinWarning("warning"):
+
+        min_samples = 5 if not min_samples else min_samples
+        metric = 'squared_dist_norm2' if not metric else metric
+        algorithm = 'brute' if not algorithm else algorithm
+
+        algorithm = get_algorithm_name(algorithm, BRUTE_FORCE,
+            [BRUTE_FORCE, KD_TREE], 'DBSCAN')
+
+        _validate_dbscan(schema_madlib, source_table, output_table, id_column,
+                         expr_point, eps, min_samples, metric, algorithm)
+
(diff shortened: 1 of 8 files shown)


fmcquillan99:
(1)
```
madlib=# select madlib.dbscan('usage')                                                                                        ;
                                    dbscan                                     
-------------------------------------------------------------------------------
                                                                               
 -----------------------------------------------------------------------       
                             USAGE                                             
 -----------------------------------------------------------------------       
 SELECT madlib.dbscan(                                                         
     source_table,       -- Name of the training data table                    
     output_table,       -- Name of the output table                           
     id_column,          -- Name of id column in source_table                  
     expr_point,         -- Column name or expression for data points          
     eps,                -- The minimum radius of a cluster                    
     min_samples,        -- The minimum size of a cluster                      
     metric              -- The name of the function to use to calculate the   
                         -- distance                                           
     algorithm           -- The algorithm to use for dbscan.                   
     ); 
```
Missing comma after metric


(2)
```
madlib=# select madlib.dbscan_predict('usage')                                                                                ;
ERROR:  function madlib.dbscan_predict(unknown) does not exist
LINE 1: select madlib.dbscan_predict('usage')
               ^
```  
This usage function seems to be missing


(3) Suggest make following params optional:
```
min_samples, default: 5
metric, default: 'squared_dist_norm2'
algorithm, default: 'brute_force'
```


(4) Can you give me an example of how you would do batch prediction?

orhankislal:
@fmcquillan99 Thanks for the comments. Here is an example on how to run the prediction on a table:
`select id_in, madlib.dbscan_predict('out1',data) from dbscan_train_data ;`

fmcquillan99:
(5)
To be consistent with knn, can we please make the algorithm `brute_force` as the base.  It seems you can do `brute` but `brute_force` does not work:
```
SELECT madlib.dbscan( 
				'dbscan_train_data',	-- source table
				'dbscan_result',		-- output table
				'pid',					-- point id column
				'pointsxx',				-- data point
				 1.75,					-- epsilon
				 4,						-- min samples
				 'dist_norm2',	-- metric
				'brute_force');			-- algorithm

ERROR:  plpy.Error: dbscan Error: algorithm has to be one of the following: brute, kd-tree (plpython.c:5038)
  PL/Python function "dbscan", line 21, in <module>
    return dbscan.dbscan(**globals())
  PL/Python function "dbscan", line 38, in dbscan
  PL/Python function "dbscan", line 184, in _validate_dbscan
  PL/Python function "dbscan", line 123, in _assert
PL/Python function "dbscan"
```


(6) 
It seems the column header `points` does not work, maybe in conflicts with an internal name?
I think it could be a common col name so we should fix this.
```
SELECT madlib.dbscan( 
				'dbscan_train_data',	-- source table
				'dbscan_result',		-- output table
				'pid',					-- point id column
				'points',				-- data point
				 1.75,					-- epsilon
				 4,						-- min samples
				 'dist_norm2',	-- metric
				'brute');			-- algorithm

ERROR:  plpy.SPIError: column reference "points" is ambiguous
LINE 3:             SET points = points
                                 ^
QUERY:  
            UPDATE dbscan_result AS t1
            SET points = points
            FROM dbscan_train_data AS t2
            WHERE t1.pid = t2.pid
        
  PL/Python function "dbscan", line 21, in <module>
    return dbscan.dbscan(**globals())
  PL/Python function "dbscan", line 109, in dbscan
PL/Python function "dbscan"
```


(7)
Please remove the verbose output unless you need it for debugging and such:
```
ALTER        ANALYZE      BEGIN        CHECKPOINT   CLOSE        CLUSTER      COMMENT      COMMIT       COPY         
CREATE       DEALLOCATE   DECLARE      DELETE FROM  DISCARD      DO           DROP         END          EXECUTE      
EXPLAIN      FETCH        GRANT        INSERT       LISTEN       LOAD         LOCK         MOVE         NOTIFY       
PREPARE      REASSIGN     REINDEX      RELEASE      RESET        REVOKE       ROLLBACK     SAVEPOINT    SELECT       
SET          SHOW         START        TABLE        TRUNCATE     UNLISTEN     UPDATE       VACUUM       VALUES       
WITH         ABORT     
etc.   
```

src/ports/postgres/modules/dbscan/dbscan.py_in
```diff
@@ -0,0 +1,331 @@
...
+        # Find the connections between core points to form the clusters
+        core_edge_table = unique_string(desp='core_edge_table')
+        plpy.execute("DROP TABLE IF EXISTS {0}".format(core_edge_table))
+        sql = """
+            CREATE TABLE {core_edge_table} AS
+            SELECT __src__, __dest__
+            FROM {distance_table} AS __t1__, (SELECT array_agg({id_column}) AS arr
+                                              FROM {core_points_table}) __t2__
+            WHERE __t1__.__src__ = ANY(arr) AND __t1__.__dest__ = ANY(arr)
+            {dist_src_sql}
```
reductionista, line 95: Somehow a comment I left here got dropped when I submitted.

I was just wondering, maybe we should add `AND __t1__.__src__ != __t1__.__dest__` to the WHERE clause?  It don't know that it hurts anything to have an extra 0-length edge for each core point, but maybe adds some minor extra work for wcc to do having them in there.  Or, maybe the filter should go earlier where the distance is computed instead?

orhankislal:
jenkins retest this

fmcquillan99:
LGTM

PR #498, opened 2020-05-18
DL: Add utility function for custom functions

khannaekta:
This PR adds new helper functions for loading and deleting from a
given table custom functions that can be passed to keras fit/evaluate.
The user passes in the function name along with a valid dill pickled
object. This commit also adds dev-check tests for it.

### DIFF
--- a/doc/mainpage.dox.in
+++ b/doc/mainpage.dox.in
@@ -292,6 +292,7 @@ Interface and implementation are subject to change.
     @{
         @defgroup grp_gpu_configuration GPU Configuration
         @defgroup grp_keras Keras
+        @defgroup grp_custom_function Load Custom Functions
         @defgroup grp_keras_model_arch Load Models
         @defgroup grp_model_selection Model Selection for DL
         @brief Train multiple deep learning models at the same time for model architecture search and hyperparameter selection.
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
@@ -1,0 +1,245 @@
+# coding=utf-8
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+
+import dill
+import plpy
+from plpy import spiexceptions
+from utilities.control import MinWarning
+from utilities.utilities import _assert
+from utilities.utilities import get_col_name_type_sql_string
+from utilities.validate_args import columns_missing_from_table
+from utilities.validate_args import input_tbl_valid
+from utilities.validate_args import quote_ident
+from utilities.validate_args import table_exists
+
+module_name = 'Keras Custom Function'
+class CustomFunctionSchema:
+    """Expected format of custom function table.
+       Example uses:
+
+           from utilities.validate_args import columns_missing_from_table
+           from madlib_keras_custom_function import CustomFunctionSchema
+
+           # Validate names in cols list against actual table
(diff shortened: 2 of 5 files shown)


fmcquillan99:
Preliminary comments:

(1)
in addition to 
```
delete_custom_function(
    object_table,
    id
)
```
I suggest we also support
```
delete_custom_function(
    object_table,
    name
)
```
since `name` is unique and is the way that users will pass object to Keras, so I suggest we allow to delete by `name` in addition to `id`


(2)
duplicate object name gives an error as it should, but it is not a user friendly error: 
```
InternalError_: AttributeError: 'module' object has no attribute 'UniqueViolation' (plpython.c:5038)
  PL/Python function "load_custom_function", line 21, in <module>
    madlib_keras_custom_function.load_custom_function(**globals())
  PL/Python function "load_custom_function", line 42, in wrapper
  PL/Python function "load_custom_function", line 93, in load_custom_function
PL/Python function "load_custom_function"
```
Could we trap this error and give a friendlier error message?


(3)
selection on id not contiguous:
```
SELECT id, name FROM test_custom_function_table ORDER BY id;
 id |  name   
----+---------
  1 | sum_fn
  3 | sum_fn1
  5 | sum_fn2
  ```
  Is there a reason why this is not 1,2,3?  It is contiguous with model ids for model architecture table so I would expect the same thing here, plus probably what the user would expect.


(4)
minor spelling error

```
select madlib.delete_custom_function( 'test_custom_function_table', 1);

ERROR:  plpy.Error: Keras Custom Funtion: Object id 1 not found (plpython.c:5038)
  PL/Python function "delete_custom_function", line 21, in <module>
    madlib_keras_custom_function.delete_custom_function(**globals())
  PL/Python function "delete_custom_function", line 42, in wrapper
  PL/Python function "delete_custom_function", line 121, in delete_custom_function
PL/Python function "delete_custom_function"
```
miss spelled "Funtion"

(5)
When we delete the last id in the table, the table gets deleted:
```
select madlib.delete_custom_function( 'test_custom_function_table', 1);

ERROR:  plpy.Error: Keras Custom Funtion error: Input table 'test_custom_function_table' does not exist. (plpython.c:5038)
  PL/Python function "delete_custom_function", line 21, in <module>
    madlib_keras_custom_function.delete_custom_function(**globals())
  PL/Python function "delete_custom_function", line 42, in wrapper
  PL/Python function "delete_custom_function", line 104, in delete_custom_function
  PL/Python function "delete_custom_function", line 619, in input_tbl_valid
PL/Python function "delete_custom_function"
```
Is this expected behavior?  i.e., do we do this with model arch table?  Maybe it's OK, not sure.  The other option is to not delete the table, but leave it empty.

**Comment by khannaekta** [comment]
For (1), (4): will update code for the changes

For (3), (5): These are expected behaviors aligned with model arch table load/delete helper functions. 

- For (3), since `id` is a sequence, if delete is called between loads, it will skip the deleted id
- For (5), Kept it consistent with model arch table delete function, we can keep the table empty too, but it makes more sense to delete it as it would anyways be created on calling load with the same tablename.
Below is the flow for load/delete model arch table as well as custom function table:
```
load()       -> Creates table id =1
load()       -> insert into table id =2
delete(id=2) -> delete from table id =2 
load()       -> insert into table id =3
delete(id=3) -> delete from table id =3 
delete(id=1) -> delete from table id =1 and drop table
```

For (2): When I try to load an object with same function name(on GPDB6), it fails with the following error:
I ran the following (2 times):
```
Function 'sum_fn1' already exists in new_tab
```
It's unable to catch that exception with GPDB5. They are probably different, but I will look into it a little more

fmcquillan99:
Ok, thanks for the comments @khannaekta .  I agree with your assessment:

Fix (1), (4)
Leave (3), (5) as is, OK
Figure out what's up with (2) on GPDB5

fmcquillan99:
(1)
Delete by name

{code}
SELECT madlib.delete_custom_function( 'test_custom_function_table', 'mult_fn');
INFO:  Keras Custom Function: Object id None has been deleted from test_custom_function_table.
CONTEXT:  PL/Python function "delete_custom_function"
 delete_custom_function 
------------------------
{code}
OK


(2)
better error message:

{code}
InternalError: (psycopg2.errors.InternalError_) plpy.Error: Function 'sum_fn' already exists in test_custom_function_table (plpython.c:5038)
  PL/Python function "load_custom_function", line 21, in <module>
    madlib_keras_custom_function.load_custom_function(**globals())
  PL/Python function "load_custom_function", line 42, in wrapper
  PL/Python function "load_custom_function", line 98, in load_custom_function
PL/Python function "load_custom_function"

[SQL: SELECT madlib.load_custom_function('test_custom_function_table', custom_function_object_sum(), 'sum_fn', 
                                   'returns sum');]
(Background on this error at: http://sqlalche.me/e/2j85)
{code}
OK

LGTM

PR #504, opened 2020-07-13
DBSCAN: Add indexing optimizations to improve the runtime

orhankislal:
The brute force DBSCAN runs on N^2 time. To improve this,
we added two layers of indexing.
1. The data is broken into chunks with kd-tree
2. Each leaf of the kd-tree creates an rtree index for
efficient range queries.

In addition we added a separate process to reduce the number
of edges to consider during wcc operation.


orhankislal:
jenkins retest please

PR #514, opened 2020-08-25
DL: Improve the predict output with prob

orhankislal:
JIRA: MADLIB-1451

The prob output for predict created a bunch of columns for different
class values. This commit pivots this format so the table signature
is the same but the number of rows is increased thanks to a separate
row for each class value.

In addition, this commit adds the option to filter the output by
a min probability or top n results as well as a ranking column for 
non-response options.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras.sql_in
@@ -1869,7 +1869,7 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.madlib_keras_predict(
     id_col                  VARCHAR,
     independent_varname     VARCHAR,
     output_table            VARCHAR,
-    pred_type               VARCHAR DEFAULT 'all',
+    pred_type               VARCHAR DEFAULT 'prob',
     use_gpus                BOOLEAN DEFAULT FALSE,
     mst_key                 INTEGER DEFAULT NULL
 ) RETURNS VOID AS $$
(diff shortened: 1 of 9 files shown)


fmcquillan99:
updated user doc as:

```
pred_type (optional)
TEXT, default: 'prob'. Type of output desired, where 'prob' gives predictions for all classes and their associated 
probabilities.  To see only the top probability use 'response'.  Alternatively, you can specify a filter for top n or 
minimum probability.  For top n, use  an INTEGER > 0 to indicate the top ranked probabilities to output.  
For minimum probability, specify a REAL value between 0.0 and 1.0 for the cutoff.
```

'prob' (default). -- used to be 'all' please change back to 'prob'

'response' -- don't show rank col

updated user doc

INTEGER 1, 2, 3... yes show rank col

DOUBLE between 0.0 and 1.0 yes show rank col

differentiate between 1.0 and 1

suggested error message

plpy.Error: madlib_keras_predict: Invalid value for pred_type param (azdadg). Must be integer>0, double precision between 0.0 and 1.0, 'response' or 'prob'. (plpython.c:5038)

fmcquillan99:
LGTM

reductionista:
We should update the usage message (lines 384-394 in `madlib_keras_predict.py_in`:
```
The output table ('output_table' above) contains the following columns:	The output table ('output_table' above) contains the following columns:
id:                 Gives the 'id' for each prediction, corresponding	id:                 Gives the 'id' for each prediction, corresponding
                    to each row from the test_table.	                    to each row from the test_table.
estimated_COL_NAME: (For pred_type='response') The estimated class for	estimated_COL_NAME: (For pred_type='response') The estimated class for
                    classification, where COL_NAME is the name of the	                    classification, where COL_NAME is the name of the
                    column to be predicted from test data.	                    column to be predicted from test data.
prob_CLASS:         (For pred_type='prob' for classification) The	prob_CLASS:         (For pred_type='prob' for classification) The
                    probability of a given class. There will be one	                    probability of a given class. There will be one
                    column for each class in the training data.	                    column for each class in the training data.
                    TODO change this
```
But other than that looks good!

PR #517, opened 2020-09-18
DL: [AutoML] Set plan_cache_mode when calling fit multiple model

khannaekta:
JIRA: MADLIB-1447

There is a known issue when calling fit multiple model with huge model
weights(JIRA: MADLIB-1414). The fix for that issue was to set
`plan_cache_mode` guc in gpdb6 to `force_generic_plan`. For automl,
since we directly call madlib_keras_fit_multiple_model from the within
the automl function, we need to set the GUC plan_cache_mode when calling
it.


reductionista:
LGTM

PR #520, opened 2020-10-01
DL: Restrict access to the custom functions table

orhankislal:
This commit ensures that any custom function table is cretated in
the madlib schema. This table should be accessible by the admin only,
which means the load_custom_function is only available to admins or
superusers as well.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
@@ -21,9 +21,13 @@ import plpy
 from utilities.control import MinWarning
 from utilities.utilities import _assert
 from utilities.utilities import get_col_name_type_sql_string
+from utilities.utilities import current_user
+from utilities.utilities import is_superuser
+from utilities.utilities import get_schema
 from utilities.validate_args import columns_missing_from_table
 from utilities.validate_args import input_tbl_valid
 from utilities.validate_args import quote_ident
+from utilities.validate_args import unquote_ident
 from utilities.validate_args import table_exists
 
 module_name = 'Keras Custom Function'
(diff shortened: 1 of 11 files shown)


src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
```diff
@@ -61,49 +61,58 @@ def _validate_object(object, **kwargs):
...
+
         plpy.execute(insert_query,[name, description, object], 0)
     # spiexceptions.UniqueViolation is only supported for PG>=9.2. For
     # GP5(based of PG8.4) it cannot be used. Therefore, checking exception
     # message for duplicate key error.
     except Exception as e:
         if 'duplicate key' in e.message:
             plpy.error("Function '{0}' already exists in {1}".format(name, object_table))
+        elif 'permission denied' in e.message:
+            plpy.error("This function can only be used by a superuser.".format(name, object_table))
```
reductionista, line 105: I guess technically they just need INSERT or CREATE permission on the madlib schema.  But hopefully they will be equivalent, assuming the admin doesn't mess with the permissions on it.

I can't think of a reason why they would, but we should probably warn them in the docs not to give CREATE or INSERT permission on madlib schema to anyone besides a superuser, since it will give them equivalent privileges.

src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.py_in
```diff
@@ -254,13 +261,17 @@ class MstSearch():
...
 
+        self.schema_madlib = schema_madlib
         self.model_arch_table = model_arch_table
         self.model_selection_table = model_selection_table
         self.model_selection_summary_table = add_postfix(
             model_selection_table, "_summary")
         self.model_id_list = sorted(list(set(model_id_list)))
 
+        if object_table is not None and schema_madlib not in object_table:
+            object_table = "{0}.{1}".format(schema_madlib, object_table)
```
reductionista, line 272: This also looks insecure.
Let's add this test as well:

  `object_table="my_insecure_schema.madlib_vulnerability"`

src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
```diff
@@ -61,49 +61,58 @@ def _validate_object(object, **kwargs):
...
 
     plpy.info("{0}: Added function {1} to {2} table".
               format(module_name, name, object_table))
 
 @MinWarning("error")
-def delete_custom_function(object_table, id=None, name=None, **kwargs):
-    object_table = quote_ident(object_table)
+def delete_custom_function(schema_madlib, object_table, id=None, name=None, **kwargs):
+
+    if object_table is not None and schema_madlib not in object_table:
```
reductionista, line 114: ^^

reductionista (changes requested):
I see two different ways in which an ordinary user could load arbitrary python code from their own schema and execute it.  We should add both of these examples (see comments) as tests and get them passing,

src/ports/postgres/modules/deep_learning/madlib_keras_custom_function.py_in
```diff
@@ -68,6 +72,8 @@ def load_custom_function(schema_madlib, object_table, object, name, description=
     _validate_object(object)
     _assert(name is not None,
             "{0}: function name cannot be NULL!".format(module_name))
+    _assert(is_superuser(current_user()), "DL: The user has to have admin "\
+        "privilages to load a custom function")
```
reductionista, line 76: This is great!

src/ports/postgres/modules/utilities/utilities.py_in
```diff
@@ -788,6 +789,10 @@ def current_user():
     return plpy.execute("SELECT current_user")[0]['current_user']
 # ------------------------------------------------------------------------
 
+def is_superuser(user):
+
+    return plpy.execute("SELECT rolsuper FROM pg_catalog.pg_roles "\
+                        "WHERE rolname = '{0}'".format(user))[0]['rolsuper']
```
reductionista, line 795: This is great!  Glad you found an easy way to verify this.

src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.py_in
```diff
@@ -268,8 +268,12 @@ class MstSearch():
...
         self.model_id_list = sorted(list(set(model_id_list)))
 
-        if object_table is not None and schema_madlib not in object_table:
-            object_table = "{0}.{1}".format(schema_madlib, object_table)
+        if object_table is not None:
+            schema_name = get_schema(object_table)
+            if schema_name is None:
+                object_table = "{0}.{1}".format(schema_madlib, quote_ident(object_table))
+            elif schema_name != schema_madlib:
+                plpy.error("DL: Custom function table has to be in the {0} schema".format(schema_madlib))
```
reductionista, line 277: Looks good

reductionista (approved):
LGTM

PR #526, opened 2020-12-18
DL: AutoML encapsulation

reductionista:
Following the principle of encapsulation, AutoML should keep track of its own temporary table names, instead of relying on assumptions about where they will be stored internally to the FitMultipleModel class.

These changes allow it to co-exist with the Model Hopper refactor (but are independent from that PR), and should help make AutoML more robust to any future changes we make inside FitMultipleModel.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
@@ -27,6 +27,7 @@ from madlib_keras_model_selection import ModelSelectionSchema
 from keras_model_arch_table import ModelArchSchema
 from utilities.validate_args import table_exists, drop_tables, input_tbl_valid
 from utilities.validate_args import quote_ident
+from madlib_keras_helper import DISTRIBUTION_KEY_COLNAME
 
 class AutoMLConstants:
     BRACKET = 's'
@@ -40,9 +41,11 @@ class AutoMLConstants:
     SKIP_LAST = 'skip_last'
     HYPERBAND_PARAMS = [R, ETA, SKIP_LAST]
     LOSS_METRIC = 'training_loss_final'
-    TEMP_MST_TABLE = unique_string('temp_mst_table')
-    TEMP_MST_SUMMARY_TABLE = add_postfix(TEMP_MST_TABLE, '_summary')
-    TEMP_OUTPUT_TABLE = unique_string('temp_output_table')
+    MST_TABLE = unique_string('mst_table')
+    MST_SUMMARY_TABLE = add_postfix(MST_TABLE, '_summary')
+    MODEL_OUTPUT_TABLE = unique_string('output_table')
+    MODEL_SUMMARY_TABLE = add_postfix(MODEL_OUTPUT_TABLE, '_summary')
+    MODEL_INFO_TABLE = add_postfix(MODEL_OUTPUT_TABLE, '_info')
     METRICS_ITERS = 'metrics_iters' # custom column
     NUM_CONFIGS = 'num_configs'
     NUM_ITERS = 'num_iterations'
@@ -79,6 +82,7 @@ class KerasAutoML(object):
         self.model_id_list = sorted(list(set(model_id_list)))
         self.compile_params_grid = compile_params_grid
         self.fit_params_grid = fit_params_grid
+        self.dist_key_col = DISTRIBUTION_KEY_COLNAME
 
         if object_table is not None:
             object_table = "{0}.{1}".format(schema_madlib, quote_ident(object_table))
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
```diff
@@ -172,45 +174,38 @@ class KerasAutoML(object):
...
                 SELECT
                     $MAD${self.source_table}$MAD$::TEXT AS source_table,
                     {validation_table}::TEXT AS validation_table,
                     $MAD${self.model_output_table}$MAD$::TEXT AS model,
                     $MAD${self.model_info_table}$MAD$::TEXT AS model_info,
-                    (SELECT dependent_varname FROM {model_training.model_summary_table})
-                    AS dependent_varname,
-                    (SELECT independent_varname FROM {model_training.model_summary_table})
-                    AS independent_varname,
+                    (SELECT dependent_varname FROM {a.MODEL_SUMMARY_TABLE})
```
kaknikhil, line 184: I am assuming we are creating the automl summary table by querying from the fit multiple summary table. If that's the case, I think it's better to get the fit multiple summary table name from the model_training object since that's the source of truth.
What do you think ?

kaknikhil (commented):
changes look good, left a couple of comments.

src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
```diff
@@ -79,6 +82,7 @@ class KerasAutoML(object):
         self.model_id_list = sorted(list(set(model_id_list)))
         self.compile_params_grid = compile_params_grid
         self.fit_params_grid = fit_params_grid
+        self.dist_key_col = DISTRIBUTION_KEY_COLNAME
```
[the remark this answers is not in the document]
reductionista: Currently it's not required, but while working on the Model Hopper refactor I realized it would help a lot with warm start if we eventually do require it.

As I was working on optimizing weight initialization, I realized if we could rely on model output tables always having a dist key, then that would speed things up and avoid unnecessary work.  Otherwise the first step has to be copying the table over to one which does have the dist key, which usually involves shuffling the weights around to different segments.  If there is no dist key, then we can't assume anything about how the weights are distributed so there is no way to optimize that part.   For all we (or gpdb) knows, all of the weights might be on the same segment with none on any other segments.

All newly generated output tables will have the dist key in them (I should make that change to fit also, come to think of it), but because they won't for v1.17 I don't require it as an input for warm start yet... we still do the extra unnecessary shuffling each time for backwards compatibility.

So there's nothing necessary about it right now, but the earlier we get this into the codebase the earlier we can drop compatibility for warm start on output tables missing a dist key.

src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
reductionista: My intention with this PR was to try to avoid as much as possible having AutoML depend on the names and meanings of internal class variables in FitMultiple.  In the case of the model output table, there were multiple different table names inside this case, and while refactoring they changed around.  So the motivation was mainly to avoid the headache that resulted from having to go through and modify all the references to those variables in AutoML again, if they change in the future.

The name of the model output table is chosen by AutoML rather than by FitMultiple, so the was a strong case for having AutoML manage that one.  For the summary table, I agree that the case is weaker... since the name is not entirely chosen by AutoML... the _summary prefix is appended to whatever AutoML chooses by FitMultiple.

I considered leaving that one as-is, but decided it seems better to have AutoML add the _summary itself for a couple reasons:  first, if we do change anything in the future in FitMultiple, it's much more likely to be the class variable name that changes than the convention of adding '_summary', since that is part of the user-facing API:  user specifies what the output table name is, and they should expect _summary to be appended as described in the docs.  AutoML is using FitMultiple in a way similar to how a user would... so it will only break if we break backwards compatibility for the user... but won't break if we refactor the code inside FitMultiple.

src/ports/postgres/modules/deep_learning/madlib_keras_automl.py_in
```diff
@@ -172,45 +183,38 @@ class KerasAutoML(object):
...
                     {object_table}::TEXT AS object_table,
                     {self.use_gpus} AS use_gpus,
-                    (SELECT metrics_compute_frequency FROM {model_training.model_summary_table})::INTEGER
-                    AS metrics_compute_frequency,
-                    {name}::TEXT AS name,
-                    {descr}::TEXT AS description,
+                    (SELECT metrics_compute_frequency FROM {a.MODEL_SUMMARY_TABLE})::INTEGER 
+                        AS metrics_compute_frequency,
+                    $MAD${self.name}$MAD$::TEXT AS name,
+                    $MAD${self.description}$MAD$::TEXT AS description,
```
khannaekta, line 207:
```suggestion
                    {name}::TEXT AS name,
                    {descr}::TEXT AS description,
```
As we have the code to check for NULL and add `$MAD$` in L180-181.

PR #532, opened 2021-01-21
DL: Add multiple variable support

orhankislal:
This commit adds support for multiple dependent and independent variables.
These changes should allow the users to use YOLO v3 and v4 models.
Note that the existing interface with single dependent and single independent
variables still works as expected. The output table formats have changed
slightly to accomodate this new feature. The summary tables store various
fields such as num_classes or dep_vartype as arrays, even if they have single
entries.

The implementation reads up to 5 dependent and 5 independent variables
separately thanks to the new interface. If more variables are passed, they are
packed into bytea arrays. A high number of large variables might cause us to go
over the 1 GB limit of Postgres due to packing.

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
@@ -29,8 +29,10 @@ from utilities.utilities import add_postfix
 from utilities.utilities import get_segments_per_host
 from utilities.utilities import unique_string
 from utilities.utilities import get_psql_type
+from utilities.utilities import split_quoted_delimited_str
 from utilities.validate_args import get_expr_type
 from utilities.validate_args import input_tbl_valid
+from utilities.validate_args import quote_ident
 
 from madlib_keras_wrapper import *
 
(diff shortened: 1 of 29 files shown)


src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -86,18 +87,34 @@ class BasePredict():
...
+        full_class_name_list = []
+        full_class_value_list = []
+
+        for i in range(self.dependent_var_count):
+
+            if self.pred_vartype[i] in ['text', 'character varying', 'varchar']:
+
+                unnest_sql.append("unnest(ARRAY{0}) AS {1}".format(
+                    ['NULL' if j is None else j for j in class_values[i]],
+                    self.dependent_varname[i]))
```
reductionista, line 101: I think this should be:
   quote_ident(self.dependent_varname[i])
instead of:
   self.dependent_varname[i]

src/ports/postgres/modules/deep_learning/madlib_keras_predict.py_in
```diff
@@ -116,63 +133,82 @@ class BasePredict():
...
-                 {self.dependent_varname} {self.pred_vartype},
+                 class_name TEXT,
+                 class_value TEXT,
                  {pred_col_name} {pred_col_type},
                  rank INTEGER)
-            """.format(**locals()))
+            """.format(**locals())
+        plpy.execute(sql)
+
+        independent_varname_sql = ["{0}::REAL[]".format(i) for i in self.independent_varname]
```
reductionista, line 146: quote_ident(self.independent_varname)

reductionista:
LGTM

PR #537, opened 2021-02-02
Fix some C++11 issues

reductionista:
This is an assortment of bug fixes and improvements for our optional C++11 support.

It's mostly only relevant for those who were compiling by passing the -DCXX11 option manually.
Previously, that was the only way to compile with C++11, and it only affected clang (not gcc).
But in addition to fixing some bugs and warnings, this PR also improves the logic in the CMake
configuration files.  Previously if you had too recent a version of boost ( > v1.64) and the -DCXX11
option was not passed, it would just fail.  With the new behavior, it will auto-enable CXX11 support
if the version of Boost requires it.  It could still fail, if the compiler doesn't support C++11, but this
is no worse than the previous behavior.  If the option is passed manually and the detected Boost is
too old, it will still fail as before... but with a hint to upgrade Boost before trying again.
Also, if boost is not detected, the version it automatically downloads will now depend on whether 
-DCXX11 was passed.

This has still only been tested on OSX platform, but the option is now available for other platforms as
well (in case it works).

Critical bug fixes:

    - Use bernoulli_distribution from boost::, not stl:: or boost::TR1

    The TR1 sub namespace no longer exists in recent versions of boost.
    Previously, this was fixed by using the C++11 STL version if it's
    available, otherwise falling back on TR1.  But that didn't work
    (caused a crash) due to a slight incompatibility between the C++11 and
    the boost version.  The boost version still exists, just not under TR1--and
    we already use it elsewhere in the codebase, so this should be
    the cleanest fix.

    - Fix memory-corruption crash in std::domain_error(), by copying formatted
      string to memory allocated safely via postgres with palloc()

Non-critical fixes/improvements:

    - Convert ptr_fun() to function() to get rid of C++11 deprecation warning
     
    - src/dbal/Reference_proto.hpp:  Fix an error in the template definition
      of class Ref.  The default value for IsMutable should be the value true
      or false, not the type bool.  The only reason this wasn't generating
      compiler errors is that IsMutable happens to be passed explicitly in
      every place in the code where the template is instantiated--so the
      default was ignored.

    - Hide global boost symbols from external libraries
      We were already doing this for all STL symbols--for the same reasons,
      doing it for boost as well should make MADlib more robust when
      integrating with 3rd party libraries.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -34,21 +34,25 @@ set(EIGEN_BASE_URL
 # running cmake:
 # "-DBOOST_TAR_SOURCE=/path/to/boost_x_x_x.tar.gz"
 
-set(BOOST_TAR_VERSION "1.61.0")
-set(BOOST_TAR_MD5 874805ba2e2ee415b1877ef3297bf8ad)
-
-string(REPLACE "." "_" _BOOST_TAR_VERSION_UNDERSCORES ${BOOST_TAR_VERSION})
-set(BOOST_TAR "boost_${_BOOST_TAR_VERSION_UNDERSCORES}.tar.gz")
-set(BOOST_URL "${SOURCEFORGE_BASE_URL}/boost/files/${BOOST_TAR}")
+if (CXX11)
+    set(BOOST_TAR_VERSION "1.75.0")  # If CXX11 is enabled, we can use latest version of Boost
+    set(BOOST_TAR_MD5 38813f6feb40387dfe90160debd71251)
+else(CXX11)
+    set(BOOST_TAR_VERSION "1.61.0")   # Most recent version that includes TR1 (necessary for old C++ compilers)
+    set(BOOST_TAR_MD5 874805ba2e2ee415b1877ef3297bf8ad)
+endif(CXX11)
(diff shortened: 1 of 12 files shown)


src/CMakeLists.txt
```diff
@@ -34,22 +34,28 @@ set(EIGEN_BASE_URL
...
+set(BOOST_CXX11_TAR "boost_${_BOOST_CXX11_TAR_VERSION_UNDERSCORES}.tar.gz")
+set(BOOST_TR1_URL "${SOURCEFORGE_BASE_URL}/boost/files/${BOOST_TR1_TAR}")
+set(BOOST_CXX11_URL "${SOURCEFORGE_BASE_URL}/boost/files/${BOOST_CXX11_TAR}")
 
 if(NOT BOOST_TAR_SOURCE)
-    find_file(BOOST_TAR_SOURCE ${BOOST_TAR}
+    find_file(BOOST_TAR_SOURCE ${BOOST_TR1_TAR}
         PATHS ${MAD_THIRD_PARTY}/downloads)
+    if(CXX11)
+        set(BOOST_TAR_SOURCE ${BOOST_TR1_URL})
```
kaknikhil, line 53: shouldn't this be `BOOST_CXX11_URL` and the else should be `BOOST_TR1_URL` ?

kaknikhil:
another comment

PR #539, opened 2021-02-08
Kmeans: Fix auto kmeans tests

orhankislal:
The auto kmeans dev check tests were using the incorrect column
for expr_point which caused some flaky tests.


khannaekta (approved):
LGTM

PR #551, opened 2021-02-23
Fix boost version bug in cmake < 3.15

kaknikhil:


reductionista:
LGTM

PR #552, opened 2021-02-24
DL: Remove with clause while running tf operations

kaknikhil:
JIRA: MADLIB-1467

We noticed that we run tf operations like `model.fit` on gpu device
using `with tf.device(/gpu:0):`. It turns out that some of the
operations(like passing class_weight as a fit param) may not be fully
supported on gpus. Since we already set `CUDA_VISIBLE_DEVICES` env
variable, we don't explicitly need to use the with clause to force the
operation to be run on gpus. So to fix this issue, this commit removes all
the with clauses so that operations are allowed to run on cpu as well.

Co-authored-by: Ekta Khanna <ekhanna@vmware.com>


orhankislal (approved):
LGTM

PR #553, opened 2021-02-24
DL: Remove AOcontrol from model selection functions

orhankislal:

### DIFF
--- a/src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.sql_in
+++ b/src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.sql_in
@@ -44,14 +44,14 @@ Interface and implementation are subject to change. </em>
 <li class="level1"><a href="#related">Related Topics</a></li>
 </ul></div>
 
-This module generates model configurations 
+This module generates model configurations
 for training multiple models at the same time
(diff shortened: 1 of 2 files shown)


src/ports/postgres/modules/deep_learning/madlib_keras_model_selection.sql_in
```diff
@@ -873,9 +872,9 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.generate_model_configs(
     object_table            VARCHAR DEFAULT NULL
 ) RETURNS VOID AS $$
     PythonFunctionBodyOnly(`deep_learning', `madlib_keras_model_selection')
-    with AOControl(False):
```
kaknikhil, line 876: If the user has appendonly set to true, then our code will fail if we don't have the `with aocontrol(False)` code. 
You can reproduce this by setting `set gp_default_storage_options='appendonly=true';` and then running load_model_selection_table/generate_model_configs will error out with
```
ERROR:  spiexceptions.FeatureNotSupported: append-only tables do not support unique indexes
```

khannaekta (approved):
LGTM!
Just a minor comment: if you could add some details on what the issue was in the commit message, it would help when we refer to this specific change in future.
Thanks!

PR #555, opened 2021-02-26
Add support for TensorBoard

orhankislal:
Co-authored-by: Domino Valdano <dvaldano@pivotal.io>

Only accepts safe options to pass to TensorBoard().  No
other callbacks allowed for now.


orhankislal:
This PR is created to squash the commits and rebase on master while keeping the old commit messages just in case.

PR #556, opened 2021-02-26
Avoid using optimization -O2 or higher if C++11 is enabled

reductionista:
There are some crashes we're seeing related to memory management when the C++11 flag is enabled.  It only happens when compiler optimizations at level O2 or higher are enabled, which makes it very difficult to debug.

Since this is not an essential feature, and currently only works on OSX anyway, in the meantime we will simply disable compiler optimization for that case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ if (CXX11)
     SET(CMAKE_CXX_FLAGS_RELEASE "-O1 -DNDEBUG"
         CACHE STRING
         "Flags used by the CXX compiler during RELEASE builds." FORCE)
-    SET(CMAKE_CXX_FLAGS_DEBUG  "-O0 -g"
+    SET(CMAKE_CXX_FLAGS_DEBUG  "-Og -g"
         CACHE STRING
         "Flags used by the CXX compiler during DEBUG builds." FORCE)
     SET(CMAKE_CXX_FLAGS_MINSIZEREL "-Os -DNDEBUG"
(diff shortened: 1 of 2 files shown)


reductionista:
Without the cmake flag set, it passes all install-check and dev-check tests on these platforms:

CentOS 6.10 / gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23) / cmake 3.5.2
CentOS 6.10 / (GCC) 6.2.0, 64-bit / cmake 3.5.2
Centos 7 / (GCC) 6.2.0, 64-bit / cmake 3.5.2
Ubuntu 16 / gcc 6.2.0 / cmake 3.5.2
Ubuntu 18 / gcc 6.2.0 / cmake 3.5.2
MacOS Catalina / gcc  / cmake 
MacOS Big Sur / gcc (Homebrew GCC 8.2.0) 8.2.0 / cmake 3.18.4

With CXX11=1 set, it compiles and passes most of the install-check and dev-check tests on these platforms:

CentOS 6.10 / gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23) / cmake 3.5.2
CentOS 6.10 / (GCC) 6.2.0, 64-bit / cmake 3.5.2
Centos 7 / (GCC) 6.2.0, 64-bit / cmake 3.5.2
Ubuntu 16 / gcc 6.2.0 / cmake 3.5.2
Ubuntu 18 / gcc 6.2.0 / cmake 3.5.2
MacOS Big Sur / Apple clang version 12.0.0 (clang-1200.0.32.29) / cmake 3.18.4

C++11 support is still "experimental" so we're okay for now with some tests failing for C++11 in kmeans, linalg, and recursive_partition.  Lowering the optimization helps get rid of some of them, but not all.

PR #573, opened 2021-11-05
@orhankislal WCC: Optimize subtx count and catalog entry frequency

orhankislal:
WCC had a high number of plpy.execute commands. Since each call
creates a new subtransaction, it was constantly hitting the
overflow limit. This commit merges most of them.

WCC also created and dropped temp tables in each iteration.
This bloats the catalog, so this commit uses a few actual
tables and uses truncate/insert.

### DIFF
--- a/src/ports/postgres/modules/graph/wcc.py_in
+++ b/src/ports/postgres/modules/graph/wcc.py_in
@@ -121,7 +121,6 @@ def wcc(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
     new_to_update_where_condition = ''
     edge_to_update_where_condition = ''
     edge_inverse_to_update_where_condition = ''
-    grouping_cols_select = ''
 
     BIGINT_MAX = 9223372036854775807
     component_id = 'component_id'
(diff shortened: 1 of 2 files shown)


src/ports/postgres/modules/graph/wcc.py_in
```diff
@@ -162,52 +160,128 @@ def wcc(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
-        plpy.execute("""
-                CREATE TEMP TABLE {message} AS
-                SELECT {vertex_id},
-                        CAST({vertex_id} AS BIGINT) AS {component_id}
-                        {select_grouping_cols_clause}
-                FROM {newupdate}
-                {distribution}
-            """.format(select_grouping_cols_clause=comma_grouping_cols,
-                       **locals()))
+
```
reductionista, line 165: Can we move this definition of `group_by_clause_newupdate` into the grouping_cols specific conditional block below?

src/ports/postgres/modules/graph/wcc.py_in
```diff
@@ -299,24 +315,27 @@ def wcc(schema_madlib, vertex_table, vertex_id, edge_table, edge_args,
...
+    plpy.execute("""
+        INSERT INTO {out_table_summary} VALUES
+        ({grouping_cols_txt} '{vertex_table}', '{vertex_id}',
+        '{vertex_id_type}');
+
+        DROP TABLE IF EXISTS {message},{oldupdate},{newupdate},{toupdate};
+    """.format(grouping_cols_txt='' if not grouping_cols else
+                    "'{0}', ".format(grouping_cols),
+               **locals()))
```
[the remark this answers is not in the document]
orhankislal: The common problem is, on GPDB 5X, we cannot create and use a table in the same block. That is why I had to shop up a number of blocks in the prep stage. I can combine them into a single `create table ... as ...` block.

src/ports/postgres/modules/graph/wcc.py_in
orhankislal: It is already in the grouping_cols specific conditional block (starting at Line 141).

reductionista: Oh!  Sorry, read it wrong

reductionista (approved):

PR #576, opened 2022-01-03
Release v1.19.0 rc2

orhankislal:

### DIFF
--- a/deploy/Readme_OSX
+++ b/deploy/Readme_OSX
@@ -1,0 +1,24 @@
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements. See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License. You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


deploy/Readme_OSX
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+MADlib Tarball Usage for OSX
+
+This tarball is generated for using MADlib on OSX devices.
```
kaknikhil, line 18: I would suggest that we reword this a bit

1. We don't need to mention `After unpacking` since this README is only accessible after the user would have unpacked already.
1. Mention that link_madlib_folders.sh needs to be run from the directory that it's located in. Also add a warning that if it's run from somewhere else, it will not work and that it might have unintended side effects like a random bin file getting deleted

kaknikhil (approved):
LGTM. Left a minor comment. Also it would be good to add more information to the commit message.

PR #577, opened 2022-03-01
GLM-multinom: Use non-temp tables in GroupIterationController

orhankislal:
There is a potential issue with pg_temp not cleaning up correctly in
the case of a failure in GLM and multinom. This commit changes the
default value to create the temp table used for state aggregation to
avoid the temporary tables.

### DIFF
--- a/src/ports/postgres/modules/glm/glm.py_in
+++ b/src/ports/postgres/modules/glm/glm.py_in
@@ -195,7 +195,8 @@ def __glm_compute(schema_madlib, tbl_source, tbl_output, col_dep_var, col_ind_va
             'rel_state': unique_string(),
             'col_grp_iteration': unique_string(),
             'col_grp_state': unique_string(),
-            'state_type': schema_madlib + ".bytea8"
+            'state_type': schema_madlib + ".bytea8",
+            'temporaryTables': False
             }
     args.update(optim_params)
     args.update(family_params)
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/glm/glm.py_in
```diff
@@ -350,7 +351,7 @@ def __glm_compute(schema_madlib, tbl_source, tbl_output, col_dep_var, col_ind_va
                    **args))
 
     # clean up
-    plpy.execute("""DROP TABLE IF EXISTS pg_temp.{rel_state} """.format(**args))
```
khannaekta, line 353: Do we call this only once? Since this isn't defined as a temp table, we probably would want to reclaim the space of table, similar to the DL module using `dev_opt_unsafe_truncate_in_subtransaction` guc. (https://github.com/apache/madlib/blob/master/src/ports/postgres/modules/deep_learning/madlib_keras_fit_multiple_model.py_in#L985-L988)

khannaekta (commented):
- Also, we will need to update the `GroupIterationController`, which creates TEMP tables too iteratively and is called here. https://github.com/apache/madlib/blob/master/src/ports/postgres/modules/utilities/in_mem_group_control.py_in#L292
- We can probably add the test, which would be just failing once and then rerunning the glm query after deleting the tables. It probably seems weird to add such a test in install-check but will leave it to you if it makes sense to add such a test.

src/ports/postgres/modules/glm/glm.py_in
orhankislal: Yes this is called only once at the end of the training.

khannaekta (approved):
LGTM!

PR #580, opened 2022-05-02
Update documentation and disable ORCA for various modules

orhankislal:


src/ports/postgres/modules/convex/mlp.sql_in
```diff
@@ -1346,99 +1349,6 @@ JOIN mlp_regress_prediction USING (id);
 (1 row)
 </pre>
 
-<h4>Regression with Grouping and Mini-Batching</h4>
```
khannaekta, line 1349: Why did we remove this?

tool/docker_start.sh
```diff
@@ -83,9 +83,10 @@ docker exec "${CONTAINER_NAME}" bash -c "rm -rf /madlib/build_docker; \
...
-docker exec "${CONTAINER_NAME}" bash -c "/madlib/build_docker/src/bin/madpack -p postgres \
-							 -c postgres/postgres@localhost:5432/postgres \
-							 install" \
+docker exec "${CONTAINER_NAME}" bash -c "PATH=$PATH:/usr/lib/postgresql/10/bin \
+							/madlib/build_docker/src/bin/madpack -p postgres \
+							-c postgres/postgres@localhost:5432/postgres \
+							install" \
 			| tee "${workdir}/build_docker_logs/madlib_install.log"
 
 docker exec "${CONTAINER_NAME}" bash -c "/madlib/build_docker/src/bin/madpack -p postgres \
```
khannaekta, line 92: Do we also need to update the PATH here for install-check?

khannaekta (commented):
LGTM! Just minor comments.

tool/docker_start.sh
orhankislal: No, only the install step needs pg_ctl in the path. Install check or any other madlib function call is fine without it.

src/ports/postgres/modules/convex/mlp.sql_in
orhankislal: 1- A bit redundant, we already have examples for:
- MLP with grouping
- MLP with minibatching
- minibatching with grouping in the minibatching docs
2- With such a small dataset, it was occasionally failing on my local mac setup.

khannaekta (approved):

PR #582, opened 2022-06-09
Utilities: Reuse update plan in GroupIterationController 

orhankislal:
The group iteration controller prepares a plan and executes it during
the update phase. For some modules, this plan does not change between
iterations. With this commit, the plan gets saved and reused as needed.

Co-authored-by: Bhuvnesh Chaudhary <bchaudhary@pivotal.io>

### DIFF
--- a/src/ports/postgres/modules/utilities/in_mem_group_control.py_in
+++ b/src/ports/postgres/modules/utilities/in_mem_group_control.py_in
@@ -240,6 +240,7 @@ class GroupIterationController:
         self.finished_states = state_factory(self.is_state_type_bytea8)
 
         self.group_param = self._init_group_param()
+        self.update_plan = None
 
     def _init_group_param(self):
         _grp_key = ("array_to_string(ARRAY[{grouping_str}], ',')"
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/utilities/in_mem_group_control.py_in
```diff
@@ -549,40 +550,43 @@ class GroupIterationController:
...
+                newState=newState,
+                iteration=self.iteration,
+                using_str=group_param.using_str,
+                groupby_str=group_param.groupby_str,
+                _grp_key=group_param.grp_key,
+                select_rel_state=group_param.select_rel_state,
+                select_n_tuples=group_param.select_n_tuples,
+                **self.kwargs)
+
+            self.update_plan = plpy.prepare(run_sql,
```
khannaekta, line 580: For `plpy.prepare()` statement, don't we want to pass along the `self.iteration` everytime? With initializing it just once for `self.update_plan` will we still be doing that?

orhankislal: That is what we used to do. Now, we keep it in memory (instead of the table) attached to the results dictionary on line 589.

khannaekta (approved):
LGTM!

PR #590, opened 2022-10-13
XGBoost: Fix class label verification in predict

orhankislal:
- Fix class label verification in predict

Class label is an optional argument for predict but the code checks to
ensure the column exists in the table. This commit fixes the issue and
adds a test to run predict with default values and no optional
parameters.

- Add support for bigint and varchar id col

XGBoost supports non-integer values as id columns (not features) in the
python implementation. This commit alters the surrounding code to
accomodate for such column types and adds/alters tests accordingly.

- Add eval_metrics as a parameter

eval_metrics is used by XGBoost used monitoring the training result and
early stopping. We expose this parameter to the user and parse it to
pass it to the fit function (instead of init).


khannaekta (approved):
LGTM!

PR #592, opened 2023-02-13
SSSP: Fix negative cycle check

orhankislal:
SSSP checks the iteration counter to identify negative cycles. If a shortest path has the same length as the number of vertices, then the check incorrectly identifies this as a negative cycle. This commit fixes the issue and adds a relevant test.

### DIFF
--- a/src/ports/postgres/modules/graph/sssp.py_in
+++ b/src/ports/postgres/modules/graph/sssp.py_in
@@ -325,7 +325,10 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
                     WHERE {vertex_id} IS NOT NULL
                     """.format(**locals()))[0]['count']
 
-            for i in range(0, v_cnt + 1):
+            # The algorithm should converge in less than |V|+1 iterations.
+            # Otherwise there is a negative cycle in the graph.
+            # We check for v_cnt+2 because the first iteration is spent on the setup
+            for i in range(0, v_cnt + 2):
 
                 # Apply the updates calculated in the last iteration.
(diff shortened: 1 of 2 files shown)


src/ports/postgres/modules/graph/sssp.py_in
```diff
@@ -325,7 +325,10 @@ def graph_sssp(schema_madlib, vertex_table, vertex_id, edge_table,
                     WHERE {vertex_id} IS NOT NULL
                     """.format(**locals()))[0]['count']
 
-            for i in range(0, v_cnt + 1):
+            # The algorithm should converge in less than |V|+1 iterations.
+            # Otherwise there is a negative cycle in the graph.
+            # We check for v_cnt+1 because the first iteration is spent on the setup
```
kaknikhil, line 330: shouldn't this comment be `We check for v_cnt+2`

src/ports/postgres/modules/graph/test/sssp.sql_in
```diff
@@ -280,9 +280,46 @@ INSERT INTO e VALUES
 (ascii('s'), ascii('t'), 6),
 (ascii('s'), ascii('y'), 7);
 
-select graph_sssp('v',
+CREATE TABLE e_grp(src INT, dest INT, weight float8, grp INT);
```
kaknikhil, line 283: I think might be a good idea to add a test case for an actual negative cycle to make sure we didn’t add any regression for detecting negative cycles

kaknikhil (approved):
LGTM. Left a minor comment about adding a testcase

PR #593, opened 2023-02-13
Disable TestIfNoUTF8BOM.py for all platforms

orhankislal:


chrishajas (approved):
Looks good, can we also add the reason

PR #602, opened 2023-05-17
Build: Add PG13 on Ubuntu 20 support for Jenkins

orhankislal:
This commit fixes the Jenkins pipeline for PG13. It disables some tests (with FIXME tag) since they had docker specific memory issues. PMML is also fixed for py3.
Finally, it adds PG14 and PG15 support for future convenience.

### DIFF
--- a/tool/jenkins/jenkins_build.sh
+++ b/tool/jenkins/jenkins_build.sh
@@ -51,35 +51,45 @@ docker pull madlib/postgres_13:jenkins
 # Launch docker container with volume mounted from workdir
 echo "-------------------------------"
 cat <<EOF
-docker run -d -e POSTGRES_PASSWORD=postgres --name madlib -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
+docker run -d -t -e POSTGRES_PASSWORD=postgres --name madlib --ulimit core=-1  --privileged -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
 EOF
-docker run -d -e POSTGRES_PASSWORD=postgres --name madlib -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
+docker run -d -t -e POSTGRES_PASSWORD=postgres --name madlib --ulimit core=-1  --privileged -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
 echo "-------------------------------"
 
 ## This sleep is required since it takes a couple of seconds for the docker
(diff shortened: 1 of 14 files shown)


tool/docker/base/Dockerfile_postgres_11_Jenkins
```diff
@@ -16,25 +16,26 @@
...
-                       python-pip \
-                       python-dev \
+                       python3-pip \
+                       python3-dev \
                        build-essential \
                        cmake \
                        libspatialindex-dev
 
-RUN pip install tensorflow==1.14 dill "rtree>=0.8,<0.9"
+RUN python3 -m pip install tensorflow dill rtree xgboost mock pandas numpy scikit-learn
```
khannaekta, line 38: don't we need the specific versions anymore?

tool/jenkins/jenkins_build.sh
```diff
@@ -51,35 +51,46 @@ docker pull madlib/postgres_13:jenkins
 # Launch docker container with volume mounted from workdir
 echo "-------------------------------"
 cat <<EOF
-docker run -d -e POSTGRES_PASSWORD=postgres --name madlib -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
+docker run -d -t -e POSTGRES_PASSWORD=postgres --name madlib --ulimit core=-1  --privileged -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
 EOF
-docker run -d -e POSTGRES_PASSWORD=postgres --name madlib -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
+docker run -d -t -e POSTGRES_PASSWORD=postgres --name madlib --ulimit core=-1  --privileged -v "${workdir}":/madlib madlib/postgres_13:jenkins | tee logs/docker_setup.log
+# -c log_statement=all -c logging_collector=on -c log_directory=logs -c log_filename=postgresql.log
```
khannaekta: we can probably remove this commented line

tool/docker/base/Dockerfile_postgres_11_Jenkins
orhankislal: With python3 the latest versions work as expected. If we start to get any issues, we should get an error and try to fix it, or at least pin it to a different version.

tool/jenkins/jenkins_build.sh
orhankislal: They were useful while I was trying to collect postgres logs. I'll add a comment to mention that.

khannaekta (approved):

PR #603, opened 2023-05-19
Build: Add gppkg v2 support for GP7

orhankislal:

### DIFF
--- a/src/ports/greenplum/cmake/FindGreenplum_6.cmake
+++ b/src/ports/greenplum/cmake/FindGreenplum_6.cmake
@@ -17,3 +17,64 @@
 
 set(_FIND_PACKAGE_FILE "${CMAKE_CURRENT_LIST_FILE}")
 include("${CMAKE_CURRENT_LIST_DIR}/FindGreenplum.cmake")
+
+function(add_gppkg GPDB_VERSION GPDB_VARIANT GPDB_VARIANT_SHORT UPGRADE_SUPPORT)
+    string(TOLOWER ${GPDB_VERSION} GPDB_VERSION_LC)
+    string(REPLACE "." "_" VERSION_ "${GPDB_VERSION}")
+
+    # Get information about the rhel version
+    rh_version(RH_VERSION)
+    string(REGEX MATCH "([0-9])" RH_MAJOR_VERSION "${RH_VERSION}")
+
+    file(WRITE "${CMAKE_BINARY_DIR}/deploy/gppkg/${GPDB_VARIANT}_${VERSION_}_gppkg.cmake" "
+    file(MAKE_DIRECTORY
+        \"\${CMAKE_CURRENT_BINARY_DIR}/${GPDB_VERSION}/BUILD\"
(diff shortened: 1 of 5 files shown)


src/ports/greenplum/cmake/FindGreenplum_6.cmake
```diff
@@ -17,3 +17,64 @@
...
+    configure_file(
+        madlib.spec.in
+        \"\${CMAKE_CURRENT_BINARY_DIR}/${GPDB_VERSION}/SPECS/madlib.spec\"
+    )
+    configure_file(
+        gppkg_spec.yml.in
+        \"\${CMAKE_CURRENT_BINARY_DIR}/${GPDB_VERSION}/gppkg/gppkg_spec.yml\"
+    )
+
+    if(GPPKG_BINARY AND RPMBUILD_BINARY)
```
khannaekta, line 54: Would it be helpful to create a separate file `FindGreenplum_7.cmake`?

src/ports/postgres/modules/bayes/test/bayes.sql_in
```diff
@@ -392,5 +392,17 @@ $$ language plpgsql;
...
-SELECT install_test_5();
+
+-- FIXME: ORCA started to fail with GP7 beta3 while accessing bayes created views in these tests
+m4_include(`SQLCommon.m4')
+m4_changequote(`<!', `!>')
+m4_ifdef(<!__POSTGRESQL__!>, <!!>, <!
+SET optimizer = off;
+!>)
+
+ SELECT install_test_4();
```
khannaekta, line 403: do these test fail with gppkg v2 installation only or with GP7 beta3 in general ?

src/ports/greenplum/cmake/FindGreenplum_6.cmake
orhankislal: This PR is for the madlib2-master so the focus is on GP7+ & PG13+. GP6 has py3 capabilities via an extension so it is deemphasised.

src/ports/postgres/modules/bayes/test/bayes.sql_in
orhankislal: I believe they fail with beta3. Even if we compile from source, we still fail these tests. I don't see any connection between ORCA optimizer failing to create a plan (even `explain` crashes the db) and gppkg.

src/ports/greenplum/cmake/FindGreenplum_6.cmake
khannaekta: Thanks! Was just suggesting it as we have a separate `.cmake` for each version.

khannaekta (approved):
LGTM!

src/ports/greenplum/cmake/FindGreenplum_6.cmake
orhankislal: We do have `FindGreenplum_7.cmake` in the same folder (in `madlib2-master` branch). It was added in a previous commit that added gp7 support.

khannaekta: Just realized, I was looking at the main branch.

PR #607, opened 2023-08-17
Various: Reduce SERIAL usage

orhankislal:
Since SERIAL columns do not guarantee a sequence of numbers with no gaps, this commit removes its usage from assoc_rules and load_keras_model.
The issue was unearthed by a change in the GPDB code setting the cache for sequences to 20 and causing the numbers to jump unexpectedly. While we can set the cache to 1 to get the previous behavior, it is better to remove the dependency.

### DIFF
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
@@ -322,6 +322,7 @@ def assoc_rules(madlib_schema, support, confidence, tid_col,
             )
          m4_ifdef(`__POSTGRESQL__', `', `DISTRIBUTED BY (id)')
          """.format(**locals()));
+    plpy.execute("ALTER SEQUENCE {assoc_loop_aux}_id_seq CACHE 1".format(**locals()));
 
     if verbose  :
         plpy.info("{0} Frequent itemsets found in this iteration".format(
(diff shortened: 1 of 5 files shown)


src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -380,29 +380,33 @@ def assoc_rules(madlib_schema, support, confidence, tid_col,
...
-                  {assoc_rule_sets_loop_ordered} t3
-             WHERE t1.newrownum < t3.newrownum AND
-                   t3.newrownum-t1.newrownum <= {num_products_threshold}
-           ) t
-           WHERE {madlib_schema}.svec_l1norm(set_list)::INT = {iterp1} AND
-                 {madlib_schema}.svec_l1norm(tids)::FLOAT8 >= {min_supp_tranx}
+
+        sql = """
+            INSERT INTO {assoc_loop_aux}(id, set_list, support, tids)
+            SELECT row_number() OVER (), *
```
khannaekta: I understand the reason for adding `row_number()` here but the only concern I have is that it would generate a plan that will gather data on the coordinator which may have performance slow down with growing data.
we can probably rewrite this query to add rownumbers without using this window function(similar to how we do it for DL's minibatch preprocessor commit: 85eba2968b5651b6242a398df569b1f3f6412703) but that would make this too complicated. Would like to know if using serial datatype and setting the cache to 1 seems better over the performance.

khannaekta (approved):
Thanks @orhankislal! LGTM

PR #615, opened 2024-03-02
PMML: Update user docs

kaknikhil:
Starting 0cd28f9733927d63beaefc9488db7f8bfdb3bd80, we no longer include intercept as a predictor in the pmml file. User docs need to be updated to incorporate these changes

This commit makes the following changes to the user docs:
1. Remove the predictor variable "1" from the namespec expression
2. Add a note about non array independent variable expressions

### DIFF
--- a/src/ports/postgres/modules/pmml/table_to_pmml.sql_in
+++ b/src/ports/postgres/modules/pmml/table_to_pmml.sql_in
@@ -111,31 +111,29 @@ SELECT madlib.pmml('patients_logregr');
 Result:
 <pre class="result">
 <?xml version="1.0" standalone="yes"?>
-<PMML version="4.1" xmlns="http://www.dmg.org/pmml-v4-1.html">
+<PMML version="4.1" xmlns="http://www.dmg.org/PMML-4_1">
   <Header copyright="redacted for this example">
     <Extension extender="MADlib" name="user" value="gpadmin"/>
-    <Application name="MADlib" version="1.7"/>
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/pmml/table_to_pmml.sql_in
```diff
@@ -146,10 +144,25 @@ Alternatively, the above can also be invoked as below if custom names are needed
...
+
+-- Now use the columns 'x' and 'y' created in the previous step
+SELECT madlib.logregr_train(
+        'patients2',
+        'patients_logregr2',
+        'y',
+        'x');
+</pre>
+In such scenarios, the pmml code always assumes that the intercept variable "1," was already included in the independent variable
+expression. If it is not included, the exported PMML would be incorrect.
```
khannaekta, line 165: will reword this to : `If it is not included, the exported PMML would not explicitly capture it as intercept.`

kaknikhil: > If it is not included, the exported PMML would not explicitly capture it as intercept

Sure but I would still mention that the exported PMML would be incorrect. This is because we would assume that x1's coefficient is the intercept's coefficient and we would have one less predictor

khannaekta: I see, sure lets keep it as is in that case.

khannaekta (approved):

